PR #1, opened 2024-10-28
PR #1: First implementation of network simulator

AntoxaBarin:
TODO:
- Replace sliding window with Congestion window (CWND)
- Calculate package travel time


project_1/include/simulator.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <vector>
+#include <cstdint>
+#include <queue>
+
+class NetworkSimulator {
+private:
+    std::uint32_t bandwidth = 52428800;
+    std::uint32_t data_transmission_frequency_bytes = 52428800;
+    std::uint32_t data_transmission_frequency_packages = 51200;
+    std::uint32_t package_size = 1024;
```
drgulevich, line 14: nit: package -> packet

`package` usually refers to a software package. In the context of networks this is called `packet`.

project_1/src/utils.cpp
```diff
@@ -0,0 +1,60 @@
...
+    return std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+}
+
+ServerBase::ServerBase(std::uint32_t id) : id(id) { }
+
+std::uint32_t ServerBase::GetID() const {
+    return id;
+}
+
+ServerSender::ServerSender(std::uint32_t id, std::uint32_t distance_ms) : ServerBase(id), distance_μs(distance_ms) { }
```
drgulevich, line 13: Initialization `distance_μs(distance_ms)` looks strange: μs = microsecond, ms = millisecond (1000 μs).

project_1/src/simulator.cpp
```diff
@@ -0,0 +1,64 @@
...
+        // Create packages, add them to priority_queue
+        for (std::uint32_t i = 0; i < test_packages_number; ++i) {
+            auto current_time_μs = GetCurrentTimeInMicroseconds();
+            packages.emplace(server.GetID(), server_receiver.GetID(), current_time_μs, i);
+        }
+        std::cout << events.back();
+    }
+
+    auto current_time_μs = GetCurrentTimeInMicroseconds();
+    std::uint32_t sliding_window_duration_μs = 1'000'000;
```
drgulevich, line 36: Just for interest, how did you manage to type `μs`? Still, I suggest not to use special characters like `μ` and replace it by standard Latin letters to avoid possible font issues and better portability. It is common to use `us` notation for `μs` (as `u` looks similar to `μ`).

project_1/src/simulator.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    // Each server send 5 packages
+    for (ServerSender& server: servers) {
+        // Create event
+        events.emplace_back(GenerateNewID(events), server, SEND_DATA, test_packages_number);
+
+        // Create packages, add them to priority_queue
+        for (std::uint32_t i = 0; i < test_packages_number; ++i) {
+            auto current_time_μs = GetCurrentTimeInMicroseconds();
+            packages.emplace(server.GetID(), server_receiver.GetID(), current_time_μs, i);
```
drgulevich, line 30: The distances to the receiver are not used yet?

project_1/src/utils.cpp
```diff
@@ -0,0 +1,60 @@
+#include "utils.hpp"
+
+std::uint64_t GetCurrentTimeInMicroseconds() {
+    return std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
```
drgulevich, line 4: Any advantage to use the internal clock time rather than introducing an artificial (simulated) time?

project_1/src/simulator.cpp
```diff
@@ -0,0 +1,64 @@
...
+NetworkSimulator::NetworkSimulator(const std::vector<std::uint32_t>& distances_to_receiver) : servers_number(distances_to_receiver.size()) {
+    for (std::uint32_t distance: distances_to_receiver) {
+        AddServer(distance);
+    }
+    std::uint32_t server_receiver_id = 0;
+    server_receiver = ServerBase(server_receiver_id);
+}
+
+void NetworkSimulator::AddServer(std::uint32_t distance_to_receiver) {
+    if (servers.size() < servers_number) {
```
drgulevich, line 14: Do we really need `servers_number`? It looks like `servers.size()` is sufficient to get the number of servers?

project_1/include/utils.hpp
```diff
@@ -0,0 +1,76 @@
...
+    std::uint32_t GetPackageIndex() const;
+    std::uint64_t GetSenderID() const;
+
+    bool operator<(const PackageHeader& other) const;
+};
+
+// Struct represents event -- sending data or acknowledgement
+struct Event {
+private:
+    std::uint64_t id;
```
drgulevich, line 62: Do the events need to have an `id`? This functionality does not seem to be used at this stage.

project_1/include/utils.hpp
```diff
@@ -0,0 +1,76 @@
...
+};
+
+// Type for package
+struct PackageHeader {
+private:
+    std::uint64_t sender_id;
+    std::uint64_t receiver_id;
+    std::uint64_t sending_time;
+    std::uint32_t package_index;
+    std::uint32_t RTT;
```
drgulevich, line 45: This might be all right but it it worries me a bit that the packet knows about RTT. The RTT can only become known when the acknowledgement arrives back to the sender, after the packet is received.

project_1/src/simulator.cpp
```diff
@@ -0,0 +1,64 @@
...
+        }
+        std::cout << events.back();
+    }
+
+    auto current_time_μs = GetCurrentTimeInMicroseconds();
+    std::uint32_t sliding_window_duration_μs = 1'000'000;
+
+    // Collect packages arrived in last second (sliding_window_duration)
+    while (!packages.empty() && current_time_μs - packages.top().GetSendingTime() < sliding_window_duration_μs) {
+        sliding_window.push(packages.top());
```
drgulevich, line 40: As we discussed earlier, the sliding window should rather be replaced by the congestion window (CWND).

drgulevich (commented):
Great start! Thanks for the contribution. I attach my comments.

project_1/src/simulator.cpp
AntoxaBarin: I agree, it was a bad idea. I just copied `μ` symbol from some website, will use `u` instead.

project_1/src/utils.cpp
AntoxaBarin: I think there are no advantages, so it will be easier and faster to use simulated time.

project_1/include/utils.hpp
AntoxaBarin: I agree, packet doesn't know anything about RTT. To order packets we need to calculate approximate travel time, I should use it instead of RTT.

project_1/include/utils.hpp
AntoxaBarin: I agree, now there is no need to identify `Events`.

AntoxaBarin:
In this commit I resolved all review issues, except Event's ID (I'll do it and CWND in the next commit).

ArtyomPeshkov:
And also I think it would be nice to devide things connected to packets, events and servers to different files, just for cleaner code and more readability, but it might be a matter of taste. After all, we don't have a style guide or something

AntoxaBarin:
@ArtyomPeshkov, thank you for review! I completely agree with you. Code is getting bigger and bigger and we need to keep it readable and clean. Not sending packets by `ServerSender` is my mistake, it looks really unnatural.

I divided `utils.hpp` and `utils.cpp` into several files and move packets sending to method `ServerSender::SendPackets()` [here](https://github.com/Cloud-Storage-Team/algnet/pull/2/files#diff-c050ce25c6df6adb81baa2daf5ea0b96163026ecf5bb0f69a573cf1fa43c62adR65-R76).

project_1/include/simulator.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+    // Sorts packets by travel time
+    std::priority_queue<PacketHeader> packets{};
+    
+    // Contains all events 
+    std::vector<Event> events{};
+
+public:
+
+    explicit NetworkSimulator(const std::vector<std::uint32_t>&);
```
IDKWNTCMF, line 28: In my opinion it will be better to add parameter names in constructors, without them it is hard to deduce what is expected as constructor parameters or in which order without looking in source .cpp file
```suggestion
    explicit NetworkSimulator(const std::vector<std::uint32_t>& /* distances_to_receiver */);
```

project_1/include/utils.hpp
```diff
@@ -0,0 +1,97 @@
+#pragma once
+
+#include <cstdint>
+#include <iostream>
+#include <vector>
+
+enum event_type: std::uint8_t {
```
IDKWNTCMF, line 7: I suggest using `enum class` instead of `enum`. It will help to avoid possible errors with implicit conversion
```suggestion
enum class event_type: std::uint8_t {
```

project_1/include/utils.hpp
```diff
@@ -0,0 +1,97 @@
...
+    SEND_DATA,
+    ACKNOWLEDGEMENT,
+};
+
+// Type for packet
+struct PacketHeader {
```
IDKWNTCMF, line 13: I suggest to make fields of structures public to simplify. If we want to ensure that fields cannot be changed then `const` can be used

project_1/include/simulator.hpp
IDKWNTCMF: Point holds for all constructors

IDKWNTCMF (commented):
Hello! I have made some suggestions.

project_1/include/utils.hpp
```diff
@@ -0,0 +1,97 @@
...
+};
+
+// Type for server receiver, contains only ID
+struct ServerBase {
+protected:
+    std::uint64_t id;
+public:
+    ServerBase(std::uint32_t);
+    ServerBase() = default;
+    ~ServerBase() = default;
```
Az3git, line 53:
```suggestion
    virtual ~ServerBase() = default;
```

drgulevich: Guideline #4: A base class destructor should be either public and virtual, or protected and nonvirtual.
from http://www.gotw.ca/publications/mill18.htm

drgulevich:
@AntoxaBarin, please, reply to these comments so that we can merge this PR.

project_1/include/utils.hpp
AntoxaBarin: I think that public fields may bring us non-trivial errors in future, so it is better to keep data encapsulated.

drgulevich (approved):
LGTM

PR #2, opened 2024-11-06
PR #2: Add time-based simulation

AntoxaBarin:
Refactoring: `TransmissionUnit` is a base struct for `PacketHeader` and `ACK` (new struct, which represents acknowledgement).
New feature: simulation over a specified time interval (simualtion time measures in seconds).

Data: file `data/1_sec_queue_size.txt` is a result of 1-second simulation with 3 senders (distances = {10, 4, 20}).
File `data_visualization/main.ipynb` contains Python code for `1_sec_queue_size.txt` visualization using matplotlib.


AntoxaBarin:
Last commit in a few words:

- `utils.hpp`, `utils.cpp` divided into files: `server.cpp`, `server.hpp`, `packet.cpp`, `packet.hpp`, `event.cpp`, `event.hpp`.
- Packet sending now is a method of struct `ServerSender` (previously was method of `NetworkSimulator`).
- Increase network's bandwidth to 50 Gbps (previously was 50 Mbps, I set this value on the start of project).
- Update `data/30_sec_queue_size.txt` file. Now there is information about 30-second network simulation with 3 servers, for each iteration we collect `iteration index`, `number of packets stuck in queue`, `mean CWND size`.
- Update `data_visualization/main.ipynb`: add new plot with mean CWND size.

project_1/include/event.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <cstdint>
+#include <iostream>
+
+enum event_type: std::uint8_t {
+    SEND_DATA,
+    ACKNOWLEDGEMENT,
+};
+
+// Struct represents event -- sending data or acknowledgement
+struct Event {
```
drgulevich, line 14: A class definition should usually start with a public: section, followed by protected:, then private:.
https://google.github.io/styleguide/cppguide.html#Declaration_Order

project_1/include/server.hpp
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include "packet.hpp"
+
+#include <cstdint>
+#include <vector>
+#include <queue>
+
+// Type for server receiver, contains only ID
+struct ServerBase {
```
drgulevich, line 10: A class definition should usually start with a public: section, followed by protected:, then private:. Omit sections that would be empty.
https://google.github.io/styleguide/cppguide.html#Declaration_Order

project_1/include/server.hpp
```diff
@@ -0,0 +1,47 @@
...
+    std::uint64_t id;
+public:
+    ServerBase(std::uint32_t);
+    ServerBase() = default;
+
+    std::uint32_t GetID() const;
+};
+
+// Type for sending server, contains ID and distance to receiver
+struct ServerSender : ServerBase {
```
drgulevich, line 21: `struct` members are public by default
Please, also follow the declaration order https://google.github.io/styleguide/cppguide.html#Declaration_Order

project_1/include/simulator.hpp
```diff
@@ -1,17 +1,20 @@
...
+#include "event.hpp"
+#include "packet.hpp"
 
 #include <vector>
 #include <cstdint>
 #include <queue>
+#include <chrono>
 
 class NetworkSimulator {
 private:
```
drgulevich, line 13: Class members are private by default

project_1/src/simulator.cpp
```diff
@@ -2,60 +2,86 @@
...
         packets.pop();
     }
 }
 
+void NetworkSimulator::StartSimulation(std::uint32_t simulation_time_sec) {
+    bool is_first_iteration = true;
+    auto start_time = std::chrono::steady_clock::now();
+    std::uint32_t iteration = 0;
+
+    while (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - start_time).count() < simulation_time_sec) {
```
drgulevich, line 55: Do we still use the physical time rather then simulated time?

drgulevich (commented):
Thanks. My comments are attached.

project_1/src/simulator.cpp
AntoxaBarin: We are using physical time to measure time of simulation. It is convenient to run simulation for N seconds.

AntoxaBarin:
In two previous commits I fix all review issues:

- `virtual` destructors
- declaration order in `struct`/`class`
- named arguments in constructors
- replacing `enum` with `enum class`

project_1/src/event.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+std::ostream& operator<<(std::ostream& out, const Event& event) {
+    out << "Initiator ID: " << event.server_initiator.GetID();
+    if (event.type == event_type::SEND_DATA) {
+        out << ". Type: SEND_DATA"
+            << ". Number of packets: " << event.units_number << ".\n";
+    }
+    else {
+        out << ". Type: ACKNOWLEDGEMENT.\n";
+    }
```
m4xxx1m, line 14: Consider using `switch` operator instead

project_1/include/packet.hpp
```diff
@@ -0,0 +1,32 @@
...
+    explicit TransmissionUnit(std::uint64_t delivery_time_us);
+protected:
+    std::uint64_t estimated_delivery_time;
+};
+
+// Type for acknowledgement
+struct ACK : TransmissionUnit {
+    ACK(std::uint64_t receiver_id, std::uint64_t delivery_time_us);
+    std::uint64_t GetReceiverID() const;
+    std::uint64_t GetEstimatedDeliveryTime() const;
```
m4xxx1m, line 15: Maybe it would be better for `TransmissionUnit` to have this method instead of implementing it twice here and in `PacketHeader`?

project_1/include/server.hpp
```diff
@@ -0,0 +1,46 @@
...
+    ServerBase() = default;
+    virtual ~ServerBase() = default;
+
+    std::uint32_t GetID() const;
+protected:
+    std::uint64_t id;
+};
+
+// Type for sending server, contains ID and distance to receiver
+struct ServerSender : ServerBase {
```
m4xxx1m, line 21: In my opinion it would be better to divide `ServerBase` and `ServerSender` into two files

project_1/include/simulator.hpp
```diff
@@ -1,34 +1,20 @@
...
-
-    explicit NetworkSimulator(const std::vector<std::uint32_t>&);
+    explicit NetworkSimulator(const std::vector<std::uint32_t>& distances_to_receiver);
     ~NetworkSimulator() = default;
 
-    void StartSimulation();
+    void StartSimulation(std::uint32_t simulation_time_sec);
 
     // New ID = last given ID + 1
     template<typename T>
```
m4xxx1m, line 20: Why not just store last given ID in class?

project_1/include/server.hpp
```diff
@@ -0,0 +1,46 @@
...
+    bool HandleACKs();
+    void printACKs() const;
+    void SendPackets();
+private:
+    const std::uint32_t max_transmission_rate_packets = 6'250'000;
+    const std::uint32_t distance_us;
+    std::uint32_t current_time_us = 0;
+    std::uint32_t cwnd_size_in_packets = 1;
+
+    std::priority_queue<PacketHeader>& packet_queue;
```
m4xxx1m, line 44: I think it would be better to accept `packet_queue` as a parameter in `SendPackets` rather than store its reference in `ServerSender`

project_1/src/simulator.cpp
```diff
@@ -2,60 +2,86 @@
...
+        server.SendPackets();
     }
+}
 
+void NetworkSimulator::PacketsReceiving() {
+    std::uint64_t received_data_kb = 0;
+    std::uint32_t received_packets = 0;
     // Create acknowledgements
     while (!packets.empty()) {
+        if (received_data_kb >= bandwidth_bytes) {
```
IDKWNTCMF, line 31: Why do we compare data in kilobytes with data in bytes without any conversion?

IDKWNTCMF: The same question holds about arithmetic operations.

IDKWNTCMF (commented):
Left some feedback about the changes.

project_1/src/simulator.cpp
AntoxaBarin: `received_data_kb` -- is a typo, we operate only with bytes.

project_1/src/simulator.cpp
```diff
@@ -2,60 +2,86 @@
...
     for (std::uint32_t distance: distances_to_receiver) {
-        servers.emplace_back(GenerateNewID(servers), distance);
+        servers.emplace_back(GenerateNewID(servers), distance, packets);
     }
     std::uint32_t server_receiver_id = 0;
     server_receiver = ServerBase(server_receiver_id);
 }
 
-void NetworkSimulator::StartSimulation() {
+void NetworkSimulator::PacketsSending(bool is_first_iteration) {
```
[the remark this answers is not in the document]
AntoxaBarin: I agree that good code must avoid flags like this one, but in current implementation of packet sending this flag is necessary. If we remove it, servers will never send packets (because on every iteration `continue` will be executed).

I will try to resolve this code-design problem.

Issue #10, opened 2024-11-25
Network Simulator: create class Packet


drgulevich:
The class Packet should describe a network packet.

PR #18, opened 2024-12-01
Network Simulator: Class Packet -- draft

AntoxaBarin:


AntoxaBarin:
Files to review: (include/Packet.hpp)[https://github.com/Cloud-Storage-Team/algnet/pull/18/files#diff-cb4db0f3dbed345a30d4d68ecfa2a3730cd4662a42967e54bdeb524f61831383], (src/Packet.cpp)[https://github.com/Cloud-Storage-Team/algnet/pull/18/files#diff-48f5b315735f6916e5396c919dfea1d29ce367549ae44e0eeff3f65923cd73b0].
Issue #10.

drgulevich (commented):
Only files which close 10 should be here.

include/Packet.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * @brief Packet sending time in nanoseconds
+     */
+    std::uint64_t sending_time_ns;
+
+    /**
+     * @brief Packet delivery time in nanoseconds
+     */
+    std::uint64_t estimated_delivery_time_ns;
```
drgulevich, line 26: Could you, please, explain why is it *estimated* delivery time instead of just delivery time?

include/Packet.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <cstdint>
+
+class Packet {
+public:
+    Packet(std::uint64_t sender_id, std::uint64_t sending_time_ns, std::uint64_t delivery_time_ns);
```
drgulevich, line 7: Let us also include here information about the mount of transmitted data (in the form of number of successfully delivered bits). This info will be useful to calculate the delivery rate. Also, it is used in the BBR congestion control algorithm.

include/Packet.hpp
drgulevich: Also, I wonder should this be in a packet header at all...?

src/Packet.cpp
```diff
@@ -0,0 +1,9 @@
+#include "Packet.hpp"
+
+Packet::Packet(std::uint64_t sender_id, std::uint64_t sending_time_ns, std::uint64_t delivery_time_ns)
+    : sender_id(sender_id), sending_time_ns(sending_time_ns), estimated_delivery_time_ns(delivery_time_ns) { }
+
+// The earlier the estimated_delivery_time, the faster the packet will arrive
+bool Packet::operator<(const Packet& other) const {
+    return estimated_delivery_time_ns > other.estimated_delivery_time_ns;
```
drgulevich, line 8: This comparison is needed to sort packets in the priority queue, right? I wonder if this should be here... What about creating a separate class *Events* for this purpose?

include/Packet.hpp
AntoxaBarin: This value corresponds to the delivery time, I named it estimated because of a misunderstanding.

src/Packet.cpp
AntoxaBarin: Yes, this comparison is needed to sort packets. I agree, class `Packet` contains several unsuitable variables like delivery time. It is a good idea to wrap `Packet` into `Event`.

BogruAKVD:
Why the simulator does not support acknowledgment packets. It seems to be used in some overload control algorithms.

include/Packet.hpp
```diff
@@ -0,0 +1,20 @@
...
+class Packet {
+public:
+    explicit Packet(std::uint64_t sending_time_ns);
+    Packet() = default;
+    ~Packet() = default;
+
+    /**
+     * @brief Packet sending time in nanoseconds
+     */
+    std::uint64_t sending_time_ns;
```
drgulevich, line 14: > std::uint64_t sending_time_ns;

The variable name seems self-explanatory, so no need to accompany it with comment `Packet sending time in nanoseconds`.

include/Packet.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * @brief Number of successfully delivered bytes
+     */
+    std::uint32_t transmitted_data_size_bytes;
```
drgulevich, line 19: This also seems self-explanatory.

drgulevich (commented):
Let's remove all doxygen-style comments for now for simplicity.

drgulevich (commented):
Most importantly, should the packet header contain source and destination? E.g. https://en.wikipedia.org/wiki/IPv4#Header

drgulevich:
> Most importantly, should the packet header contain source and destination? E.g. https://en.wikipedia.org/wiki/IPv4#Header

In this context, https://github.com/Cloud-Storage-Team/algnet/pull/25 seems closer to the truth:
```
std::uint64_t source_id;
std::uint64_t destination_id;
```

drgulevich:
I mark this as draft and propose to go with PR https://github.com/Cloud-Storage-Team/algnet/pull/25 to close the issue #10.

drgulevich:
I close this PR as we go with https://github.com/Cloud-Storage-Team/algnet/pull/25

Issue #11, opened 2024-11-25
Network Simulator: implement a method of Flow  which sends packets


drgulevich:
Implement a method of the Flow class which sends packets.

PR #17, opened 2024-12-01
Network Simulator: class Flow

AntoxaBarin:
Closes Issue https://github.com/Cloud-Storage-Team/algnet/issues/9. Implements class Flow: a top-level user interface which describes a socket pair (source and destination).

PR #19, opened 2024-12-01
Network Simulator: method Flow::Send()

AntoxaBarin:
Closes issue #11


drgulevich (commented):
https://github.com/Cloud-Storage-Team/algnet/blob/main/README.md

AntoxaBarin:
> Sender: получает пакеты от Flow. При генерации пакета Flow вызывает Sender::Enqueue(packet). Sender после обработки (по истечении очереди и задержки) отправляет пакет к следующему устройству (например, Switch).

I can't understand why flow directly calls Sender::Enqueue(packet) instead of creating an event with Sender::ProcessPacket(packet) as handler.

- Flow creates `event` with Sender::ProcessPacket(packet)
- Scheduler executes `event`'s handler -- sender enqueues packet to the queue and then creates a new event (sending packet to next device) with delay.

Is it a good solution for Flow::Send?

drgulevich:
> I can't understand why flow directly calls Sender::Enqueue(packet) instead of creating an event with Sender::ProcessPacket(packet) as handler.

I think you have misunderstood the role of scheduler. The purpose of the scheduler is to order the events (arrival of packets coming from different sources). In your example, when everything happens on one server there is no need of using scheduler because we can order the issue of packets directly one by one.

src/Flow.cpp
```diff
@@ -13,4 +14,15 @@ Flow::Flow(const std::vector<std::shared_ptr<NetworkDevice>>& path, const std::v
...
     this->distances_ns = distances_ns;
     id = Flow::last_given_flow_id++;
+}
+
+void Flow::Send() {
+    std::shared_ptr<NetworkDevice> sender = path.front();
+    std::shared_ptr<NetworkDevice> receiver = path.back();
+
+    Packet p(sender->id, receiver->id, 0, false);
+    NetworkSimulator::Schedule(interval_ns, [&]() {
```
drgulevich, line 24: What is `interval_ns` and why do we need this?

src/Flow.cpp
```diff
@@ -16,4 +17,15 @@ Flow::Flow(const std::vector<std::shared_ptr<NetworkDevice>>& path,
     this->path = path;
     this->distances_ns = distances_ns;
     id = Flow::last_given_flow_id++;
+}
+
+void Flow::Send() {
+    std::shared_ptr<NetworkDevice> sender = path.front();
+    std::shared_ptr<NetworkDevice> receiver = path.back();
```
drgulevich, line 24: After the suggested changes in https://github.com/Cloud-Storage-Team/algnet/pull/17 this will not be needed anymore.

src/Flow.cpp
```diff
@@ -16,4 +17,15 @@ Flow::Flow(const std::vector<std::shared_ptr<NetworkDevice>>& path,
...
+    Packet p(sender->id, receiver->id, 0, false);
+
+    NetworkSimulator::Schedule(packet_generation_interval_ns, [this, sender, p]() {
```
drgulevich, line 27: Nit: `m_packet_generation_interval_ns` for member variables

drgulevich (approved):
Thanks. LGTM.

Issue #12, opened 2024-11-25
Network Simulator: create class Network Simulator with method Run()


drgulevich:
Implement class Network Simulator with the method Run() which runs the simulation.

PR #20, opened 2024-12-01
NetworkSimulator: Class NetworkSimulator

AntoxaBarin:
Closes issue #12.


drgulevich (commented):
Only files which close issue 12 should be here.

drgulevich (commented):
1. See https://github.com/Cloud-Storage-Team/algnet/blob/main/README.md

2. Let us *not* use Doxygen for now because this project is on the experimental stage. Also, most of the comments are redundant for self-explanatory variable names:

```
    /**
     * @brief Size of a packet in bytes.
     */
    static const std::uint32_t packet_size_bytes = 1024;
```
The meaning is clear here from the name of `packet_size_bytes`.

include/NetworkSimulator.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "EventScheduler.hpp"
```
drgulevich, line 3: Where is this file?

src/NetworkSimulator.cpp
```diff
@@ -0,0 +1,77 @@
+#include "NetworkSimulator.hpp"
+#include "Flow.hpp"
+
+void NetworkSimulator::Run() const {
+    event_scheduler = std::make_unique<EventScheduler>();
```
drgulevich, line 5: Where is `EventScheduler`?

drgulevich:
When ready please, convert this back from draft to PR and ping me for review.

AntoxaBarin:
Comment on the last commit.

`map<pair<src_id, dst_id>, time_ns> link_last_process_time_ns` -- is a map that stores time when the last packet was pushed through link in the direction from src_id to dst_id.

Previously, I used half-duplex link (send packet in only one direction at a time). Now 2 links form full-duplex communication. 

Devices use `NetworkSimulator::GetLinkLastProcessTime` and `NetworkSimulator::UpdateLinkLastProcessTime` to interact with links.

include/NetworkSimulator.hpp
```diff
@@ -0,0 +1,48 @@
...
+    static std::uint64_t Now();
+
+    std::uint64_t stop_time_ns = 1'000'000'000;
+    static inline std::unique_ptr<EventScheduler> event_scheduler;
+    static inline std::map<std::pair<std::uint32_t, std::uint32_t>, std::shared_ptr<Link>> routing_table;
+    static inline std::unordered_map<std::uint32_t, std::shared_ptr<NetworkDevice>> device_by_id;
+
+    static inline std::uint64_t current_time_ns = 0;
+    static const std::uint32_t packet_size_bytes = 1024;
+    static const std::uint64_t bandwidth_bytes = 6'250'000'000;
```
drgulevich, line 39: What is `bandwidth_bytes` and where is it used? Is bandwidth a good term here? The units of bandwidth is usually bps, not bytes.

src/NetworkSimulator.cpp
```diff
@@ -0,0 +1,53 @@
...
+    stop_time_ns = time_ns;
+}
+
+void NetworkSimulator::AddDevice(std::shared_ptr<NetworkDevice> device) {
+    device_by_id[device->id] = device;
+}
+
+void NetworkSimulator::AddLink(std::shared_ptr<Link> link, std::shared_ptr<Flow> flow) {
+    std::shared_ptr<Link> reversed_link = std::make_shared<Link>(link->destination, link->source, link->distance_ns);
+    // pair <current_device, receiver_id> --> link to next device
```
drgulevich, line 30: Is it a commented code or a comment?

include/NetworkSimulator.hpp
```diff
@@ -0,0 +1,48 @@
...
+    static void AddFlow(std::shared_ptr<Flow> flow);
+    static void Schedule(std::uint64_t delay, const std::function<void(void)>& handler);
+    static std::uint64_t Now();
+
+    std::uint64_t stop_time_ns = 1'000'000'000;
+    static inline std::unique_ptr<EventScheduler> event_scheduler;
+    static inline std::map<std::pair<std::uint32_t, std::uint32_t>, std::shared_ptr<Link>> routing_table;
+    static inline std::unordered_map<std::uint32_t, std::shared_ptr<NetworkDevice>> device_by_id;
+
+    static inline std::uint64_t current_time_ns = 0;
```
drgulevich, line 37: Which of these are `private` members? Just to remind, that private members will be prefixed with `m_`.

drgulevich (commented):
Thanks. See comments below.

include/NetworkSimulator.hpp
```diff
@@ -0,0 +1,48 @@
...
+    void StopAt(std::uint64_t time_ns);
+    static void AddDevice(std::shared_ptr<NetworkDevice> device);
+    static void AddLink(std::shared_ptr<Link> link, std::shared_ptr<Flow> flow);
+    static void AddFlow(std::shared_ptr<Flow> flow);
+    static void Schedule(std::uint64_t delay, const std::function<void(void)>& handler);
+    static std::uint64_t Now();
+
+    std::uint64_t stop_time_ns = 1'000'000'000;
+    static inline std::unique_ptr<EventScheduler> event_scheduler;
+    static inline std::map<std::pair<std::uint32_t, std::uint32_t>, std::shared_ptr<Link>> routing_table;
```
drgulevich, line 34: Yes, this is me who proposed to introduce a global `routing_table` but now it seems to me that making a *local* routing table would be a better solution because: 
1. It's structure will be simpler than the global one.
2. There will be a `default` next hop when no entry in the table is found.
3. This is how it is in the real system.

include/NetworkSimulator.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+class NetworkSimulator {
+public:
+    NetworkSimulator() = default;
+    ~NetworkSimulator() = default;
+
+    static inline bool EnableACK = true;
+
+    void Run() const;
+    void StopAt(std::uint64_t time_ns);
```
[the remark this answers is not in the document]
drgulevich: Yes, I know, you did what is written in the README, but it looks like `Run(std::uint64_t time_ns)` would be better.

include/NetworkSimulator.hpp
```diff
@@ -0,0 +1,48 @@
...
+    static void AddDevice(std::shared_ptr<NetworkDevice> device);
+    static void AddLink(std::shared_ptr<Link> link, std::shared_ptr<Flow> flow);
+    static void AddFlow(std::shared_ptr<Flow> flow);
+    static void Schedule(std::uint64_t delay, const std::function<void(void)>& handler);
```
BogruAKVD, line 29: I think the Network Simulator class should be non-static. Currently, the simulator is designed as a static class, which limits its functionality to a single simulation instance per program execution (as i understand). By converting it to a non-static class, we could allow multiple instances of Network Simulator to be created and run simultaneously for testing various network scenarios without needing to restart the program.

Issue #21, opened 2024-12-02
Network Simulator: Create a base class for a network device


drgulevich:
Create a base class for a network device which will allow to construct complex network topologies out of its objects. Presumably, the class should contain a list of neighbours in the form of the adjacency list of pointers. In this way, a route can be constructed by walking a graph with DFS or BFS. The class for a network device will be inherited by switches and computer nodes.

PR #22, opened 2024-12-02
NetworkDevice: Class NetworkDevice

AntoxaBarin:
Closes issue #21.


AntoxaBarin:
Do we still want to use priority queue as container for in-flight packets? 

I add `storage` to NetworkDevice, it is a map: ID --> list of packets. Map usage: source device add new packet to map[source_id], then destination device grab packet from map[source_id].  

But know I realize that we can remove `storage` from NetworkDevice and add more suitable containers for inherited Switch and ComputerNode (priority queue for Switch and queue for ComputerNode).

include/NetworkDevice.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @brief Network device type.
+     */
+    DeviceType type;
+
+    /**
+     * @brief Network device ID.
+     */
+    std::uint32_t id;
```
drgulevich, line 36: Do we really need an `id` or passing pointers to the objects is sufficient?

include/NetworkDevice.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @brief Network device storage.
+     *
+     * @details Device stores packets from adjacent device in corresponding vector. \n
+     * Key: adjacent device ID. Value: vector with packets.
+     */
+    std::unordered_map<std::uint32_t, std::vector<Packet>> storage;
```
drgulevich, line 44: `storage` here need clarification. What is the design idea that you implement here? What do you thinks about introducing ingress/egress queues instead?

AntoxaBarin: I think that idea of ingress/egress queues is better than `storage` idea.  `storage` -- is an ingress container, I thought it was a good way to store incoming packets, but now I realize that it will be more slower than priority queue.

We can use `priority_queue<Packet>` as ingress container and `queue<Packet>` as egress container.

include/NetworkDevice.hpp
AntoxaBarin: The `id` is currently being used by the `storage`, but without last one we can use pointers to objects instead.

include/NetworkDevice.hpp
drgulevich: > We can use priority_queue<Packet> 

As discussed elsewhere (see issue #12), `priority_queue` should be in the global event scheduler.

include/NetworkDevice.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @brief Network device type.
+     */
+    DeviceType type;
+
+    /**
+     * @brief Device's current time in nanoseconds
+     */
+    std::uint64_t current_time_ns = 0;
```
drgulevich, line 41:
```
   /**
     * @brief Network device type.
     */
    DeviceType type;
```


```
   /**
     * @brief Device's current time in nanoseconds
     */
    std::uint64_t current_time_ns = 0;
```

This is self-explanatory. Comments to self-expanatory variables name are redundant, unless we are absolutely sure we want to go the Doxygen way (probably not, at this stage of the project).

drgulevich (commented):
See https://github.com/Cloud-Storage-Team/algnet/blob/main/README.md

src/NetworkDevice.cpp
```diff
@@ -0,0 +1,19 @@
+#include "NetworkDevice.hpp"
+
+NetworkDevice::NetworkDevice(DeviceType type, double processing_delay_ns):
```
drgulevich, line 3: Why do we need `processing_delay_ns`? 
1. Usually it is small and can be neglected compared to other types of delays.
2. When it is not small, this can just be added to the propagation delay to the next hop.

src/NetworkDevice.cpp
```diff
@@ -0,0 +1,28 @@
...
+    Packet result = buffer.front();
+    buffer.pop();
+    return result;
+}
+
+bool NetworkDevice::Empty() const {
+    return buffer.empty();
+}
+
+std::shared_ptr<NetworkDevice> NetworkDevice::NextDevice() const {
```
drgulevich, line 22: As we have the global routing table, I think, NextDevice amd PrevDevice are not needed anymore? The routing table can be access when needed, without the need for NextDevice and PrevDevice. Moreover, there could be multiple `next devices`.

drgulevich:
Please, convert this to PR when ready and ping me for review.

drgulevich (commented):
Thanks.

include/NetworkDevice.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    virtual void ProcessPacket(Packet p) = 0;
+
+    void Enqueue(Packet p);
+    Packet Dequeue();
+    bool Empty() const;
+    std::shared_ptr<Link> NextLink(std::uint32_t destination_id) const;
+
+    std::uint64_t id;
+    std::queue<Packet> buffer;
```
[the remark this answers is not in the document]
AntoxaBarin: If we make these members private, we will have to add getter for each of them. Instead, we can make them `protected`, members will not be accessible from the outside, and derived classes (Sender, Switch, Receiver) will be able to use them without getters.

Code: 
```cpp
    ...
    std::uint64_t GetID() const;

protected:
    std::uint64_t m_id;
    std::queue<Packet> m_buffer;
    std::uint64_t m_processing_delay_per_packet;
    std::uint64_t m_completion_time = 0;
    inline static std::uint64_t m_last_given_device_id = 0;
```

include/NetworkDevice.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <cstdint>
+#include <queue>
+
+/* Forward declaration of Link class is necessary
+ because NetworkDevice uses Link and Link uses NetworkDevice */
+class Link;
+
+class NetworkDevice {
+public:
+    explicit NetworkDevice(std::uint64_t processing_delay_ns);
```
drgulevich, line 16: If we go with the *local* routing table as discussed in https://github.com/Cloud-Storage-Team/algnet/pull/20#discussion_r1918228715, then this is the class where it should live.

PR #25, opened 2024-12-12
Network Simulator: class Packet

### LINKED ISSUE #10: Network Simulator: create class Packet

ArtyomPeshkov:
Closes #10.


project_2/include/packet.hpp
```diff
@@ -0,0 +1,51 @@
...
+public:
+    std::uint64_t source_id;
+    std::uint64_t destination_id;
+    // TODO: use it as real sending time (passing first link maybe) in checks that
+    std::uint64_t sending_time;
+    std::uint32_t packet_index;
+    std::uint32_t size;
+
+    // 0 bit - is credit packet
+    // 2 bit - initializing connection or not
```
drgulevich, line 20: I see you introduced a flag on bit 2 for handshaking. Do we actually need handshaking for the credit-based congestion contol?

ArtyomPeshkov: Yes, when sender wants to send some data it has to setup connection with receiver with such packet
From the paper:
image
Starting and stopping credit flow (end-host): Finally, ExpressPass requires a signaling mechanism to start the credit flow at the receiver. For TCP, we piggyback credit request to either SYN and/or SYN+ACK packet depending on data availability. This allows data transmission to start immediately after the 3-way handshake. For persistent connections or UDP, we send credit requests in a minimum sized packet, but this adds a delay of an RTT.

drgulevich:
Thanks. I propose to go with this PR to close issue #10 as it seems more complete.

project_2/include/packet.hpp
drgulevich: Thanks for explaining. Yes, this is how it works in the real system. However, for the sake of simulation and modelling, can we assume that the connection is established and manage this within the Flow class? Sort of: 

Flow created => start sending credits; Flow deleted => stop sending credits. 

Although, this is not exact representation of the real system, this is simpler and might be sufficient for the purpose of congestion control simulations.

project_2/src/packet.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+void PacketHeader::SetFlag(std::uint8_t bit, std::uint8_t value) {
+    if (value == 0) {
+        flags = flags & ~((std::uint8_t)1 << bit);
+    } else if (value == 1) {
+        flags = flags | ((std::uint8_t)1 << bit);
+    }
+}
+
+bool PacketHeader::operator<(const PacketHeader& other) const {
```
drgulevich, line 25: Should this (comparing the sending times) be a responsibility of the event scheduler?

project_2/include/packet.hpp
ArtyomPeshkov: I think yes, we can assume that and it would be enougth in our case

project_2/src/packet.cpp
ArtyomPeshkov: Yes, it should, so I removed it from here now

include/packet.hpp
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @brief Construct a new Packet Header object
+     * 
+     */
+    PacketHeader(std::uint64_t source_id, std::uint64_t destination_id, std::uint64_t sending_time, std::uint32_t packet_index, std::uint32_t size);
+
+    std::uint8_t GetFlag(std::uint8_t bit) const;
+    void SetFlag(std::uint8_t bit, std::uint8_t value);
+
+#ifdef DEBUG
```
drgulevich, line 41: As discussed on the stand up meeting, let's remove the DEBUG flag and let `<<` to be defined for the logging purposes.

include/packet.hpp
```diff
@@ -0,0 +1,54 @@
...
+ */
+class PacketHeader {
+public:
+    std::uint64_t source_id;
+    std::uint64_t destination_id;
+    // TODO: use it as real sending time (passing first link maybe) in checks that
+    std::uint64_t sending_time;
+    std::uint32_t packet_index;
+    std::uint32_t size;
+    std::uint64_t RTT;
```
drgulevich, line 24: Nit: lower case `rtt`

include/packet.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+    // 0 bit - is credit packet
+    // 1 bit - is ACK
+    // 2 bit - initializing connection or not
+    std::uint8_t flags = 0;
```
drgulevich, line 29: As discussed on the stand up meeting, let's introduce `enum` here

include/packet.hpp
```diff
@@ -0,0 +1,54 @@
...
+#define PRINT_PACKET(pack)
+#endif
+
+/**
+ * @brief Network packet abstraction
+ * 
+ */
+class PacketHeader {
+public:
+    std::uint64_t source_id;
```
drgulevich, line 18: I surrender here in favour of getters. In this case, these become private members starting with prefix `m_`

include/packet.hpp
```diff
@@ -0,0 +1,54 @@
...
+    std::uint64_t RTT;
+
+    // 0 bit - is credit packet
+    // 1 bit - is ACK
+    // 2 bit - initializing connection or not
+    std::uint8_t flags = 0;
+
+    PacketHeader() = default;
+    /**
+     * @brief Construct a new Packet Header object
```
drgulevich, line 33: All this comment says is that contructor is a contructor.

drgulevich (commented):
Thanks. See my comments below.

drgulevich:
Please, ping me for review and convert this from draft to PR when ready. Thanks

PR #26, opened 2024-12-12
Simulation events

ArtyomPeshkov:


drgulevich (commented):
This is a good start, but instead of introducing `virtual perform_action` and inheriting from the abstract class `Event` a better idea could be to have a single class `Event` but passing a callback function to be executed (this is how event scheduling is implemented in NS3 simulator).

See also the new README update here https://github.com/Cloud-Storage-Team/algnet/blob/main/README.md

PR #27, opened 2024-12-12
Simulation network elements

ArtyomPeshkov:


drgulevich:
Good. This code can be used to close issues #30 and #31.

project_2/.gitignore
drgulevich, line 1: Not needed for issues https://github.com/Cloud-Storage-Team/algnet/issues/30 and https://github.com/Cloud-Storage-Team/algnet/issues/31.

project_2/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+cmake_minimum_required(VERSION 3.26)
```
drgulevich, line 1: Not needed for issues https://github.com/Cloud-Storage-Team/algnet/issues/30 and https://github.com/Cloud-Storage-Team/algnet/issues/31.

project_2/README.md
```diff
@@ -0,0 +1,38 @@
+## Проект 2: Алгоритм контроля перегрузки на основе кредитов на C++ (ExpressPass)
```
drgulevich, line 1: Not needed for issues https://github.com/Cloud-Storage-Team/algnet/issues/30 and https://github.com/Cloud-Storage-Team/algnet/issues/31.

project_2/include/link.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
```
drgulevich, line 1: If we with this PR to close issues https://github.com/Cloud-Storage-Team/algnet/issues/30 and https://github.com/Cloud-Storage-Team/algnet/issues/31 then the code should be moved to the root folder.

project_2/include/switch.hpp
```diff
@@ -0,0 +1,33 @@
...
+    std::uint64_t max_buffer_size_bit = 81920;
+    std::uint64_t current_buffer_size_bit = 0;
+    
+public:
+    /**
+     * @brief creates bottle neck in network and process packets
+     * 
+     * @param current_time_ns 
+     * @param packet 
+     * @param packets_wrapped
```
drgulevich, line 29: Nit: a common case when Doxygen doc string deviates from the actual implementation. Therefore, let us not use Doxygen at all in the `main` branch code to simplify the code maintenance.

project_2/src/link.cpp
```diff
@@ -0,0 +1,26 @@
...
+    speed(speed),
+    last_process_time_ns(last_process_time_ns)
+    {}
+
+std::shared_ptr<NetworkElement> Link::GetNextElement(std::uint64_t destination_id) {
+    return destination;
+}
+
+void Link::ReceivePacket(std::uint64_t current_time_ns, PacketHeader& packet, std::priority_queue<std::shared_ptr<Event>, std::vector<std::shared_ptr<Event>>, EventComparator>& all_events) {
+    std::uint64_t process_time = (packet.size + speed - 1) / speed;
```
drgulevich, line 22: What is this formula? Specifically, where `+ speed - 1` does this formula come from?

drgulevich: Also, do you take into account the propagation delay?

ArtyomPeshkov: Its just a way of performing division with rounding up to work with nanoseconds as an integer number
No, I'm not taking into account the propagation delay right now

drgulevich: I think, I am confused. All the quantities in the brackets have different units (we add size, speed and a constant). Can you, please, elaborate on where the formula
```
process_time = (packet.size + speed - 1) / speed
```
comes from.

include/link.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <cstdint>
+#include <queue>
+#include <memory>
+#include <unordered_map>
+
+/**
+ * @brief Network link abstraction
+ * 
+ */
+class Link : public NetworkElement {
```
drgulevich, line 14: This is issue #39.

include/network_element.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <queue>
+#include <memory>
+
+class RoutingPacket;
+
+/**
+ * @brief Abstract network element (link, switch, host, etc.)
+ * 
+ */
+class NetworkElement {
```
drgulevich, line 16: It seems to me that it would be better to define independent classes for a network device and a link separately, i.e. without inheritance from a common abstract class.

PR #32, opened 2024-12-13
Network Simulator: class Event

### LINKED ISSUE #29: Network Simulator: create class Event

AntoxaBarin:
Closes #29. Implements events -- callbacks scheduled at some specified time in future.

### DIFF
--- a/include/Event.hpp
+++ b/include/Event.hpp
@@ -1,8 +1,5 @@
 #pragma once
 
-#include "Packet.hpp"
-
-#include <memory>
 #include <cstdint>
 #include <functional>
 
--- a/include/Event.hpp
+++ b/include/Event.hpp
@@ -9,7 +9,7 @@ public:
 
     void Execute() const;
 
-    std::uint64_t m_execution_time;
+    std::uint64_t execution_time;
 
     bool operator<(const Event& other) const;
 private:
--- a/include/Event.hpp
+++ b/include/Event.hpp
@@ -7,7 +7,6 @@ class Event {
 public:
     Event(std::uint64_t time, const std::function<void()>& handler);
 
-    std::uint64_t GetTime() const;
     void Execute() const;
 
     std::uint64_t execution_time;
--- a/include/Event.hpp
+++ b/include/Event.hpp
@@ -9,8 +9,9 @@ public:
 
     void Execute() const;
 
-    std::uint64_t execution_time;
-    std::function<void()> handler;
+    std::uint64_t m_execution_time;
 
     bool operator<(const Event& other) const;
+private:
+    std::function<void()> m_handler;
 };
--- a/include/Event.hpp
+++ b/include/Event.hpp
@@ -4,39 +4,17 @@
 
 #include <memory>
 #include <cstdint>
+#include <functional>
 
 class Event {
 public:
-    explicit Event(std::shared_ptr<Packet> packet, std::uint32_t flow_id, std::uint32_t receiver_id,
-                   std::uint64_t delivery_time_ns, bool is_ACK);
+    Event(std::uint64_t time, const std::function<void()>& handler);
 
-    /**
-     * @brief Pointer to the packet
-     */
-    std::shared_ptr<Packet> packet;
+    std::uint64_t GetTime() const;
+    void Execute() const;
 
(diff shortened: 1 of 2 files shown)


include/Event.hpp
```diff
@@ -0,0 +1,42 @@
...
+     */
+    std::uint64_t delivery_time_ns;
+
+    /**
+     * @brief Acknowledgement flag
+     *
+     * @details True: ACK
+     * @details False: Packet
+     */
+    bool is_ACK;
```
drgulevich, line 39: I do not think `is_ack` is a property of an *event*. Should `is_ack` be a property of a *packet* instead?

drgulevich (commented):
See https://github.com/Cloud-Storage-Team/algnet/blob/main/README.md

include/Event.hpp
AntoxaBarin: Yes, it should be a property of a `Packet`.

AntoxaBarin: Yes, it should be a property of a `Packet`.

include/Event.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <functional>
+
+class Event {
+public:
+    Event(std::uint64_t time, const std::function<void()>& handler);
+
+    std::uint64_t GetTime() const;
+    void Execute() const;
+
+    std::uint64_t execution_time;
```
drgulevich, line 16: We either need a public getter `GetTime` and private `execution_time` OR public `execution_time`, but not both: if you use `GetTime` then  `execution_time` should be private.

include/Event.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "Packet.hpp"
+
+#include <memory>
```
drgulevich, line 5: Which functions use this include file?

include/Event.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "Packet.hpp"
```
drgulevich, line 3: Where do you use  "Packet.hpp"?

AntoxaBarin: Now Event and Packet are independent classes, this include is left over from the previous Event implementation.

include/Event.hpp
```diff
@@ -0,0 +1,16 @@
...
+#include <functional>
+
+class Event {
+public:
+    Event(std::uint64_t time, const std::function<void()>& handler);
+
+    void Execute() const;
+
+    std::uint64_t execution_time;
+    std::function<void()> handler;
```
drgulevich, line 13: 1. Nit: m_handler
2. *m_handler* should be private

include/Event.hpp
```diff
@@ -0,0 +1,17 @@
...
+#include <cstdint>
+#include <functional>
+
+class Event {
+public:
+    Event(std::uint64_t time, const std::function<void()>& handler);
+
+    void Execute() const;
+
+    std::uint64_t m_execution_time;
```
drgulevich, line 12: Nit: `m_` for private members only to distinguish from public data (https://lefticus.gitbooks.io/cpp-best-practices/content/03-Style.html).

drgulevich, line 12:
```suggestion
    std::uint64_t execution_time;
```

src/Event.cpp
```diff
@@ -0,0 +1,13 @@
...
+Event::Event(std::uint64_t time, const std::function<void()> &handler):
+        m_execution_time(time),
+        m_handler(handler) { }
+
+void Event::Execute() const {
+    m_handler();
+}
+
+bool Event::operator<(const Event& other) const {
+    return m_execution_time > other.m_execution_time;
```
drgulevich, line 12:
```suggestion
    return execution_time > other.execution_time;
```

drgulevich (approved):
Thanks

src/Event.cpp
```diff
@@ -0,0 +1,13 @@
+#include "Event.hpp"
+
+Event::Event(std::uint64_t time, const std::function<void()> &handler):
+        m_execution_time(time),
```
drgulevich, line 4:
```suggestion
        execution_time(time),
```

PR #33, opened 2024-12-13
Network Simulator: Class Switch

### LINKED ISSUE #30: Network Simulator: create class Switch
Switch should be derived from class NetworkDevice

AntoxaBarin:
Closes #30.


drgulevich (commented):
This seems a bit too complex. The switch should decide where to forward a packet based on the source, destination (stored in the packet) and a *routing table*. For now, let us assume that there is a universal *routing table* accessible by all network devices.

drgulevich:
PR https://github.com/Cloud-Storage-Team/algnet/pull/27 introduces network elements with a routing table. I suppose something on these lines should be here.

AntoxaBarin:
Global routing table (in case of several flows with same source and destination devices) can be `map<pair<src_id, dst_id>, map<current_device_id, vector<shared_ptr<NetworkDevice>>>>`.

Usage example:
```cpp
Packet p = buffer.front(); // buffer is a device queue 
/*
Get next device on the path.

There is a choice among all possible next devices on the path to receiver.
For example, we can pick device with the smallest queue size.

Here we choose first device in the vector of all possible next devices.
*/
shared_ptr<NetworkDevice> next_device = NetworkSimulator::routing_table[{p.m_source_id, p.m_destination_id}][id][0];
```

image

If there are no flows with same src and dst then every vector in the routing table will have only one element.

drgulevich:
`map<pair<src_id, dst_id>, map<current_device_id, vector<shared_ptr<NetworkDevice>>>>`
`map<dst_id, map<current_device_id, vector<shared_ptr<NetworkDevice>>>> (removed src_id)`

Both of these look very complex to me. All we need to know is what is the next hop, given the current device:

```
next_hop (current_device)
```

This will be just

```
std::unordered_map< std::string, std::string> routingTable;
```

There could be several diverging flows passing through one network device, but let's postpone this problem for now (this should be done in the Project 9 on load balancing).

include/Switch.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "NetworkDevice.hpp"
+#include "Packet.hpp"
+
+class Switch: public NetworkDevice {
+public:
+    Switch(double processing_delay_ns);
```
BogruAKVD, line 8: I don't know how to write about this correctly, but I want to ask you. Up to this point, all the numbers we used were integers. So, why did you decide to calculate the delays in double

src/Switch.cpp
```diff
@@ -0,0 +1,51 @@
...
+    if (p.m_is_ack) {
+        next_device = PrevDevice();
+    }
+    else {
+        next_device = NextDevice();
+    }
+
+    std::uint64_t latency = 0;
+    /* Queueing delay */
+    if (next_processing_time_ns > NetworkSimulator::Now()) {
```
drgulevich, line 22: What is `next_processing_time_ns` and where is it defined?!

src/Switch.cpp
```diff
@@ -0,0 +1,36 @@
...
+        link = NextLink(p.m_source_id);
+    }
+    else {
+        /* send packet to receiver */
+        link = NextLink(p.m_destination_id);
+    }
+    std::shared_ptr<NetworkDevice> next_device = link->destination;
+
+    /* Latency is a sum of processing and queuing delays */
+    std::uint64_t latency = processing_delay_per_packet + std::max<std::uint64_t>(0, completion_time - NetworkSimulator::Now());
```
drgulevich, line 24: `m_processing_delay_per_packet`?

src/Switch.cpp
```diff
@@ -0,0 +1,36 @@
...
+    /* Waiting for the link to process previous packets, then add distance (i.e. link transmission time) */
+    std::uint64_t transmission_delay = (std::uint64_t)std::max<std::uint64_t>(0, (std::int64_t)(link->last_process_time_ns - (NetworkSimulator::Now() + latency))) +
```
drgulevich, line 26: Should this be delegated to the link?

drgulevich (commented):
Thanks. See comments attached.

src/Switch.cpp
```diff
@@ -0,0 +1,36 @@
...
+                                       link->distance_ns;
+    /* Update link last process time */
+    link->UpdateLastProcessTime(NetworkSimulator::Now() + latency + transmission_delay);
+
+    NetworkSimulator::Schedule(latency + transmission_delay, [this, next_device]() {
```
drgulevich, line 31: I am not sure that `transmission_delay` should enter here: because `Dequeue()` will be called after `latency + transmission_delay` the order of packets can be changed and could results in sending a packet to a wrong link.

PR #34, opened 2024-12-13
Network Simulator: Class Server

AntoxaBarin:
Closes #31.

PR #40, opened 2024-12-24
Network simulator: class Link

### LINKED ISSUE #39: Network Simulator: create a data structure which describes links

AntoxaBarin:
Closes #39

Link has pointers to source and destination devices, last process time.

Variable `speed` may be usefull for some projects.


drgulevich (commented):
Thank you. See my comment here: https://github.com/Cloud-Storage-Team/algnet/pull/34#discussion_r1897906403

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "NetworkDevice.hpp"
+
+#include <cstdint>
+#include <memory>
+
+class Link {
+public:
+    Link(std::shared_ptr<NetworkDevice> source, std::shared_ptr<NetworkDevice> destination, std::uint64_t distance_ns);
+
+    void UpdateLastProcessTime(std::uint64_t new_process_time_ns);
```
txgk, line 12: A more grammatically correct name for a variable here is `new_processing_time_ns`.

```suggestion
    void UpdateLastProcessTime(std::uint64_t new_processing_time_ns);
```

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+    std::shared_ptr<NetworkDevice> source;
```
txgk, line 14: See my suggestion above.

```suggestion
    std::shared_ptr<NetworkDevice> src;
```

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+    std::shared_ptr<NetworkDevice> destination;
```
txgk, line 15: See my suggestion above.

```suggestion
    std::shared_ptr<NetworkDevice> dest;
```

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "NetworkDevice.hpp"
+
+#include <cstdint>
+#include <memory>
+
+class Link {
+public:
+    Link(std::shared_ptr<NetworkDevice> source, std::shared_ptr<NetworkDevice> destination, std::uint64_t distance_ns);
```
txgk, line 10: I'd suggest to use `latency_ns` instead of `distance_ns`.

The term "latency" specifically refers to the time delay experienced in a system, particularly in the context of data transmission. By using `distance_ns` you misleadingly assign time units to the distance which is not very technical.

```suggestion
    Link(std::shared_ptr<NetworkDevice> source, std::shared_ptr<NetworkDevice> destination, std::uint64_t latency_ns);
```

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+    void UpdateLastProcessTime(std::uint64_t new_process_time_ns);
+
+    std::shared_ptr<NetworkDevice> source;
+    std::shared_ptr<NetworkDevice> destination;
+    std::uint64_t distance_ns;
```
txgk, line 16: See my suggestion above.

```suggestion
    std::uint64_t latency_ns;
```

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+    // In bits per ns
+    std::uint64_t speed = 0;
```
txgk, line 19: Wouldn't it be better to stick to a established convention and add units of measurement to the varialbe name for self-commenting purposes? For example, `speed_bps` would be a good name. That way you don't need to add a comment and everyone using your code will have a clear understanding of what the varialbe stores straightaway.

```suggestion
    std::uint64_t speed_bps = 0;
```

txgk, line 19: I'd suggest to use `bandwidth` word instead of `speed`. Here's an explanation: `bandwidth` specifically refers to the maximum rate of data transfer measured in bits per second, accurately reflects the capacity of a network link, distinguishes itself from latency - which refers to the time delay in data transmission - emphasizes the relevance of data capacity for application performance, and **is widely used in networking literature and discussions**.

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+    // Stores time when the last packet was pushed through
+    std::uint64_t last_process_time_ns = 0;
```
txgk, line 21: A more grammatically correct name for a variable here is `last_processing_time_ns`.

```suggestion
    std::uint64_t last_processing_time_ns = 0;
```

include/Link.hpp
```diff
@@ -0,0 +1,22 @@
...
+};
```
txgk, line 22: Please, add newline to the end of file.

To avoid doing it manually every time, you can adjust your text editor to trigger adding newline to the end of file automatically on every save.

src/Link.cpp
```diff
@@ -0,0 +1,14 @@
+#include "Link.hpp"
+
+#include <utility>
+#include <vector>
+#include <memory>
+
+Link::Link(std::shared_ptr<NetworkDevice> source, std::shared_ptr<NetworkDevice> destination, std::uint64_t distance_ns):
```
txgk, line 7: See my suggestion above.

```suggestion
Link::Link(std::shared_ptr<NetworkDevice> src, std::shared_ptr<NetworkDevice> dest, std::uint64_t latency_ns):
```

src/Link.cpp
```diff
@@ -0,0 +1,14 @@
...
+        source(std::move(source)),
+        destination(std::move(destination)),
+        distance_ns(distance_ns) { }
+
+void Link::UpdateLastProcessTime(std::uint64_t new_process_time_ns) {
+    last_process_time_ns = std::max(last_process_time_ns, new_process_time_ns);
```
txgk, line 13: See my suggestion above.

```suggestion
    last_process_time_ns = std::max(last_processing_time_ns, new_processing_time_ns);
```

include/Link.hpp
drgulevich: Thanks. I do not see a big problem with `distance_ns` as we do measure distances in the light years in astronomy, but I would suggest `delay_ns`.

PR #49, opened 2025-01-31
Network Simulator: API

drgulevich:
Closes #48. API for network simulator: a refactored version of https://github.com/rcomrad/KusServer/tree/simulator/simulator.

The following model assumptions are used: 

1. The logical *Link* comprises egress queue of the source node and ingress queue of the destination device.
2. The logical *Device* redistributes packets between the ingress/egress queues.

Types of events:

1. Generation of a packet.
2. A packet arrives to the ingress queue of the physical switch (located in the logical input link).
3. A packet is dequeued (in the order given by a simple round-robin scheduling) from one of the ingress queues for processing.
4. Stop simulation.

sketch

events

PR #66, opened 2025-02-07
FEAT: Scheduler

### LINKED ISSUE #52: Network Simulator: implementation of the Scheduler class methods

ArtyomPeshkov:
Closes #52


PaulRalnikov:
Please, do `git rebase api` and resolve conflicts

drgulevich:
> Please, do `git rebase api` and resolve conflicts

I used to do merge (`git merge api`), but this is a matter of taste

source/scheduler.cpp
```diff
@@ -0,0 +1,23 @@
...
+namespace sim {
+
+void Scheduler::tick() {
+    if (!m_events.empty()) {
+        Event* event = m_events.top();
+        m_events.pop();
+        (*event)();
+        delete event;
+    } else {
+        std::cout << "No events to process." << std::endl;
```
drgulevich, line 17: What do you think about using a logger (e.g. `spdlog` seems simple and easy to use) for this purpose?
> `logger->warn("No events to process.");`

drgulevich (commented):
Thank you! Looks good, but we also need tests here for `add`, `tick` and `clear` methods (`clear` has been introduced in the latest API update).

source/scheduler.cpp
ArtyomPeshkov: Yes, I think it will be more convenient and clear this way

PaulRalnikov:
Please, reopen your PR to `main`, not to `api`

source/scheduler.cpp
```diff
@@ -0,0 +1,27 @@
...
+    if (!m_events.empty()) {
+        auto event = m_events.top();
+        m_events.pop();
+        event();
+    } else {
+        logger->warn("No events to process.");
+    }
+}
+
+void Scheduler::add(const Event& event) { m_events.emplace(event); }
```
drgulevich, line 21: Do we want to check that the event time does not exceed the stop time? Alternatively, do we want to introduce a function which returns the last event time (`std::uint32_t pick_time() const`) instead?

drgulevich (commented):
Thanks

source/scheduler.cpp
drgulevich: These are just considerations, I am not yet sure what would be the best way to stop the simulation at a given stop time

drgulevich: > One more idea: what if we just create and add a "STOP event" to the scheduler from the beginning? The comparison to the stop time will be done naturally within the priority queue, which is nice!

Done in the new commit [e835734](https://github.com/Cloud-Storage-Team/algnet/pull/49/commits/e835734dab8b035fa31c1ed5929c52af03e99c24)

drgulevich:
LGTM. Let's merge this after https://github.com/Cloud-Storage-Team/algnet/pull/49.

drgulevich (approved):
Thanks, LGTM!

Sorry for the mess with commits from the `api` branch. I should have probably done a rebase, not merge (would this help with not inheriting the commit history from the `api` branch?).

drgulevich:
The PR description should say which issue it closes.

drgulevich:
> Please, do `git rebase api` and resolve conflicts

I think you should rebase to `api` and then change the destination branch to `main`

source/scheduler.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <spdlog/spdlog.h>
+#include <iostream>
+#include <queue>
+
+#include "event.hpp"
+
+namespace sim {
+
+void Scheduler::tick() {
```
AntoxaBarin, line 11: I think `tick()` should return boolean value to notify simulator when `m_events` is empty.

```suggestion
bool Scheduler::tick() {
```

ArtyomPeshkov: Can you provide some example when we need such behaviour?

AntoxaBarin:
```

File: source/scheduler.cpp [hunk 4]
```diff
@@ -0,0 +1,31 @@
...
+
+void Scheduler::add(const Event& event) { m_events.emplace(event); }
+
+void Scheduler::clear() {
+    m_events = std::priority_queue<Event>()
+}
+
+void Scheduler::clear() {
+    m_events = std::priority_queue<Event>()
+}
```
PaulRalnikov, line 29: Duplicates code upper

drgulevich:
1. Please, update the PR description.
2. Tests are needed for the introduced functions.

CMakeLists.txt
```diff
@@ -18,9 +18,18 @@ option(BUILD_TESTS "Build tests" ON)
...
     enable_testing()
     add_test(NAME MyTests COMMAND ${TEST_NAME})
+
+    # Test for scheduler tests
+    set(SCHEDULER_TEST "scheduler_tests")
+    file(GLOB SCHEDULER test/scheduler/*.cpp source/scheduler.cpp)
+    add_executable(${SCHEDULER_TEST} ${SCHEDULER})
+    target_include_directories(${SCHEDULER_TEST} PUBLIC source)
+    target_link_libraries(${SCHEDULER_TEST} gtest gtest_main)
+    add_test(NAME SchedulerTests COMMAND ${SCHEDULER_TEST})
```
PaulRalnikov, line 34: Your tests do not wun in CI/CD (they complies to `sheduler_tests` executable, but pipeline runs only `test_simulator`). It is my bad, I wrote test building not perfect. I think, [here](https://github.com/Cloud-Storage-Team/algnet/pull/84/files) I fix it the best way

test/scheduler/add_test.cpp
```diff
@@ -0,0 +1,20 @@
+#include <gtest/gtest.h>
```
PaulRalnikov, line 1: In all your tests you run `sim::Scheduler::get_instance().clear();`. Using Google test you may do it better: just create class `ShedulerTest` that heirs from `testing::Test` and override method `TearDown` for it. To use it, create `TEST_F(SheduleTest, test_name)` and write your test inside this function

test/scheduler/utils.hpp
```diff
@@ -0,0 +1,48 @@
+#include <gtest/gtest.h>
```
ArtyomPeshkov, line 1: #pragma once

ArtyomPeshkov, line 1:
```suggestion
#pragma once

#include <gtest/gtest.h>
```

test/scheduler/add_test.cpp
ArtyomPeshkov: Thanks for advice!

test/scheduler/add_test.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+#include "utils.hpp"
+
+int sim::CountingEvent::cnt;
```
PaulRalnikov, line 5: It would be nice to move this initialization to new file `utils.cpp` to make code more understandable

test/scheduler/utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+class TestScheduler : public testing::Test {
+public:
+    void TearDown() override{
+        sim::Scheduler::get_instance().clear();
+    };
+    void SetUp() override{};
+};
+
+namespace sim {
```
PaulRalnikov, line 19: I think it would better to use other namespace:
```suggestion
namespace test {
```

test/scheduler/utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+    std::uint32_t min_time = 1;
+    std::uint32_t max_time = static_cast<std::uint32_t>(1e9);
+
+    while ((number--) > 0) {
+        std::unique_ptr<Event> event_ptr = std::make_unique<T>(T());
+        event_ptr->time = rand() % (max_time - min_time + 1) + min_time;
+        Scheduler::get_instance().add(std::move(event_ptr));
+    }
+}
+}  // namespace sim
```
PaulRalnikov, line 59:
```suggestion
}  // namespace test
```

test/scheduler/add_test.cpp
```diff
@@ -0,0 +1,19 @@
+#include <gtest/gtest.h>
+
+#include "utils.hpp"
+
+namespace test {
+
+TEST_F(TestScheduler, AddExpectedAmountOfElements) {
+    int number_of_events = 100;
```
drgulevich, line 8: Any reason why 100 is better than smaller numbers like 3 or 5? If the tests fails for only 3 then it will fail for 100 as well. 100 will not make any difference in performance for a single test, of course, but if we have 1000s of tests then, in principle, it could.

test/scheduler/stop_test.cpp
```diff
@@ -0,0 +1,23 @@
+#include <gtest/gtest.h>
+
+#include "utils.hpp"
+
+namespace test {
+
+TEST_F(TestScheduler, StopEventWorksCorrectly) {
+    int number_of_events = 50;
```
drgulevich, line 8: Maybe, just 5 would be enough?

test/scheduler/comparator_test.cpp
```diff
@@ -0,0 +1,17 @@
+#include <gtest/gtest.h>
+
+#include "utils.hpp"
+
+namespace test {
+
+TEST_F(TestScheduler, ExpectedProcessingOrder) {
+    int number_of_events = 100;
```
drgulevich, line 8: Maybe, just 5 could be enough?

drgulevich (approved):
Great work, LGTM

PR #68, opened 2025-02-14
FEAT: Simulator

### LINKED ISSUE #51: Network Simulator: implementation of the Simulator class methods

AntoxaBarin:
Closes #51.

Simulator methods:
- `add_device(...)` creates an `shared_ptr` to the new device and inserts it into unordered_map `m_graph`.
- `add_flow(...)` creates a flow and inserts it in `m_flows`.
- `add_link(...)` creates links (src->dst, dst->src) and adds them to corresponding devices.
- `start(...)` schedules a Stop event and ticks scheduler while there are events in it.
- `recalculate_paths(...)` builds routing table for each device using BFS.


PaulRalnikov:
Please, reopen your PR  to `main`, not to 'api'

source/simulator.hpp
```diff
@@ -14,6 +14,7 @@ class Flow;
 class Simulator {
```
BogruAKVD, line 14: Seems to be a missing method for creating a flow, like
```void add_flow(std::string a_source_name, std::string a_destination_name)```
More than that maybe with argument ```std::size_t packet size ```

File: source/simulator.cpp [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+        case DeviceType::SWITCH:
+            m_graph[a_name] = new Switch();
+            break;
+        case DeviceType::RECEIVER:
+            m_graph[a_name] = new Receiver();
+            break;
+    }
+}
+
+void Simulator::add_link(Device* a_node1, Device* a_node2, int a_delay) {
```
BogruAKVD, line 36: It looks like impossible to get any valid device from m_graph

source/simulator.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+Simulator::~Simulator() {
+    for (auto& p : m_graph) {
+        delete p.second;
+    }
+}
+
+void Simulator::add_device(std::string a_name, DeviceType a_type) {
+    if (m_graph.find(a_name) != m_graph.end()) {
+        return;
```
ArtyomPeshkov, line 21: I think we might want some log message here
```suggestion
        logger->warn("Device with such name already exists.");
        return;
```

source/simulator.cpp
```diff
@@ -0,0 +1,80 @@
...
+            m_graph[a_name] = new Receiver();
+            break;
+    }
+}
+
+void Simulator::add_link(Device* a_node1, Device* a_node2, int a_delay) {
+    Link* link = new Link(a_node1, a_delay);
+    a_node2->add_inlink(link);
+    link = new Link(a_node2, a_delay);
+    a_node1->add_inlink(link);
```
ArtyomPeshkov, line 40: And what about outlinks here? If I am not mistaken we also need to them to be added

PaulRalnikov: Yes, I agree with you, here we need add_outlink. It should be added to api

source/simulator.cpp
```diff
@@ -0,0 +1,80 @@
...
+        while (!q.empty()) {
+            Device* vertex = q.front();
+            q.pop();
+            std::vector<Device*> neighbors = vertex->get_neighbors();
+            for (Device* neighbor : neighbors) {
+                path[neighbor] = vertex;
+                q.push(neighbor);
+            }
+        }
+        return path;
```
PaulRalnikov, line 58: Lets move it to separate function. Also set for visited vertexes should be added and some codestyle fixed needs:
```suggestion
    // returns map, that gives for each meet device its parent in bfs bypass tree
    std::unordered_map<Device*, Device*> bfs (Device* start_device) {
        std::unordered_map<Device*, Device*> parent;
        std::queue<Device*> queue;
        std::set<Device*> used;
        queue.push(start_device);

        while (!queue.empty()) {
            Device* device = queue.front();
            queue.pop();
            if (used.contains(device)) {
                continue;
            }
            used.insert(device);
            std::vector<Device*> neighbors = device->get_neighbors();
            for (Device* neighbor : neighbors) {
                parent[neighbor] = device;
                queue.push(neighbor);
            }
        }
        return parent;
```

source/simulator.cpp
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    for (auto& [_, src_device] : m_graph) {
+        std::unordered_map<Device*, Device*> path = bfs(src_device);
```
PaulRalnikov, line 62: Naming suggestion:
```suggestion
        std::unordered_map<Device*, Device*> parent_table = bfs(src_device);
```

source/simulator.cpp
```diff
@@ -0,0 +1,80 @@
...
+        for (auto& [_, dest_device] : m_graph) {
+            Device* next_hop = dest_device;
```
PaulRalnikov, line 64: Graph might be disconnected, so some of devices might not be reachable from `src_device`. You need to consider this:
```suggestion
            Device* next_hop = dest_device;
            if (!path.contains(dest_device)) {
                 src_device->update_rounting_table(dest_device, nullptr);
                 conitnue;
            }
```

source/simulator.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "switch.hpp"
+
+namespace sim {
+
+Simulator::Simulator() : m_scheduler(Scheduler::get_instance()) {}
+
+Simulator::~Simulator() {
+    for (auto& p : m_graph) {
+        delete p.second;
+    }
```
PaulRalnikov, line 16: Also you need to delete all added links

source/simulator.hpp
```diff
@@ -15,9 +16,11 @@ class Flow;
...
 public:
     Simulator();
+    ~Simulator();
     void add_device(std::string a_name, DeviceType a_type);
-    void add_link(Device* a_from, Device* a_to, std::uint32_t a_speed_mbps,
-                  std::uint32_t m_delay);
+    void add_link(std::string a_from_name, std::string a_to_name,
```
drgulevich, line 21: I think, the original idea of https://github.com/rcomrad/KusServer/tree/simulator/simulator was to work with pointers everywhere, not string names. Strings are only reserved for convenience of reading the output.

source/simulator.hpp
```diff
@@ -15,9 +16,11 @@ class Flow;
 class Simulator {
 public:
     Simulator();
+    ~Simulator();
```
drgulevich, line 19: Changes to the API should be suggested in https://github.com/Cloud-Storage-Team/algnet/pull/49 (already done)

source/simulator.cpp
```diff
@@ -0,0 +1,89 @@
...
+                next_hop = path[next_hop];
+            }
+            src_device->update_routing_table(
+                dest_device, src_device->get_link_to_device(next_hop));
+        }
+    }
+}
+
+void Simulator::start(std::uint32_t a_stop_time) {
+    while (m_scheduler.pick_time() <= a_stop_time) {
```
drgulevich, line 82: This is not needed (see new updates to https://github.com/Cloud-Storage-Team/algnet/pull/49)

drgulevich:
This branch should be rebased and target `main`

source/simulator.cpp
```diff
@@ -0,0 +1,89 @@
...
+        m_scheduler.tick();
+    }
+}
+
+void Simulator::clear() { m_scheduler.clear(); }
```
PaulRalnikov, line 87: Might be removed

PaulRalnikov, line 87: Might be removed
```suggestion
```

PaulRalnikov:
Please, mark as resolved all completed conversations (implemented code suggestions etc)

source/simulator.cpp
AntoxaBarin: Now this method returns `Device*` and in case of name collision a `nullptr` will be returned.

PaulRalnikov:
Please, fix building of project

drgulevich:
> [Add missing cpp files](https://github.com/Cloud-Storage-Team/algnet/pull/68/commits/28feefec5607026aac1f59f2f05868d47d714c01)

Could you, please `git reset HEAD~2; git push -f`  the last 2 commits: no needs to pour everything in one PR as this becomes completely out of scope

source/device.hpp
```diff
@@ -32,30 +32,32 @@ class IRoutingDevice {
                                       std::shared_ptr<ILink> link) = 0;
     virtual std::vector<std::shared_ptr<IRoutingDevice>> get_neighbours()
         const = 0;
-    virtual std::shared_ptr<ILink> get_link_to_destination(
-        std::shared_ptr<IRoutingDevice> device) const = 0;
     virtual std::shared_ptr<ILink> next_inlink() = 0;
+    virtual std::shared_ptr<ILink> get_link_to_destination(
```
drgulevich, line 36: No changes here?

drgulevich (commented):
Just a few passing comments. Is this still in a draft stage?

source/event.hpp
```diff
@@ -58,6 +58,7 @@ struct Process : public Event {
  */
 struct Stop : public Event {
     Stop();
+    Stop(std::uint32_t a_stop_time);
```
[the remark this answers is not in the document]
AntoxaBarin: Сonstructor `Stop()` is necessary for [this testing function](https://github.com/Cloud-Storage-Team/algnet/blob/396755c3b167c8b4762e65a6f6884151eefa751c/test/scheduler/utils.cpp#L27).

AntoxaBarin:
@ArtyomPeshkov, @BogruAKVD, @PaulRalnikov, PR is ready for review

source/simulator.cpp
```diff
@@ -0,0 +1,122 @@
...
+    m_flows.emplace_back(std::make_shared<Flow>(sender, receiver, 0.f));
+}
+
+void Simulator::add_link(std::shared_ptr<IRoutingDevice> a_from,
+                         std::shared_ptr<IRoutingDevice> a_to,
+                         std::uint32_t a_speed_mbps, std::uint32_t a_delay) {
+    m_links.emplace_back(
+        std::make_shared<Link>(a_from, a_to, a_speed_mbps, a_delay));
+    a_from->add_outlink(m_links.back());
+    a_to->add_inlink(m_links.back());
```
PaulRalnikov, line 56: Lets make code more clear for understanding:
```suggestion
    auto link = std::make_shared<Link>(a_from, a_to, a_speed_mbps, a_delay);
    m_links.emplace_back(link);
    a_from->add_outlink(link);
    a_to->add_inlink(link);
```

source/simulator.cpp
```diff
@@ -0,0 +1,122 @@
...
+        used.insert(device);
+        std::set<std::shared_ptr<ILink>> outlinks = device->get_outlinks();
+        for (std::shared_ptr<ILink> link : outlinks) {
+            if (used.find(link->get_to()) != used.end()) {
+                continue;
+            }
+            if (!parent_table.contains(link->get_to())) {
+                parent_table[link->get_to()] = link;
+            }
+            queue.push(link->get_to());
```
PaulRalnikov, line 84: Move `link->get_to` to new variable:
```suggestion
           auto next_device = link->get_to();
            if (!parent_table.contains(next_device)) {
                parent_table[next_device] = link;
            }
            queue.push(next_device);
```

source/simulator.cpp
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+    return parent_table;
+};
+
+void Simulator::recalculate_paths() {
+    for (auto& [src_device_name, src_device] : m_graph) {
+        std::unordered_map<std::shared_ptr<IRoutingDevice>,
+                           std::shared_ptr<ILink>>
+            parent_table = bfs(src_device);
```
PaulRalnikov, line 94:
```suggestion
        parent_table_t parent_table = bfs(src_device);
```

source/simulator.cpp
```diff
@@ -0,0 +1,122 @@
...
+                         std::uint32_t a_speed_mbps, std::uint32_t a_delay) {
+    m_links.emplace_back(
+        std::make_shared<Link>(a_from, a_to, a_speed_mbps, a_delay));
+    a_from->add_outlink(m_links.back());
+    a_to->add_inlink(m_links.back());
+}
+
+// returns map, that for each meet device gives link to its parent in bfs bypass
+// tree
+std::unordered_map<std::shared_ptr<IRoutingDevice>, std::shared_ptr<ILink>> bfs(
```
PaulRalnikov, line 61: Lets rewrite `bfs` and `recalculate_paths` in some new logic. I suggest bfs to return rounting table. You may calculate in on the spot: when bfs goes throw `link` connecting device `from` with device `to`, we may set `routing_table[to] = routing_table[from]` excepts case `from = start_device` : in that case `routing_table[to] = link`

test/device/add_outlink_test.cpp
```diff
@@ -0,0 +1,67 @@
+#include <gtest/gtest.h>
```
PaulRalnikov, line 1: Please, move all this tests logic to existing tests in `add_inlinks_test.cpp` and rename this file (like `add_link_test.cpp`)

test/link/utils.cpp
```diff
@@ -6,25 +6,31 @@ namespace test {
...
 
+std::set<std::shared_ptr<sim::ILink>> DeviceMock::get_outlinks() const {
+    return {};
+}
+
 std::vector<std::shared_ptr<sim::IRoutingDevice>> DeviceMock::get_neighbours()
     const {
     return std::vector<std::shared_ptr<sim::IRoutingDevice>>();
-}
+};
```
PaulRalnikov, line 27: Same here
```suggestion
}
```

test/link/utils.cpp
```diff
@@ -6,25 +6,31 @@ namespace test {
...
 
 std::shared_ptr<sim::ILink> DeviceMock::get_link_to_destination(
     std::shared_ptr<IRoutingDevice> device) const {
     return nullptr;
 }
 
-std::shared_ptr<sim::ILink> DeviceMock::next_inlink() { return nullptr; }
+std::shared_ptr<sim::ILink> DeviceMock::next_inlink() { return nullptr; };
```
PaulRalnikov, line 34: Same
```suggestion
std::shared_ptr<sim::ILink> DeviceMock::next_inlink() { return nullptr; }
```

test/link/utils.hpp
```diff
@@ -17,13 +17,15 @@ class DeviceMock : public sim::IRoutingDevice {
...
                                       std::shared_ptr<sim::ILink> link) final;
-    virtual std::shared_ptr<sim::ILink> next_inlink() final;
-    virtual std::vector<std::shared_ptr<IRoutingDevice>> get_neighbours()
+    std::vector<std::shared_ptr<sim::IRoutingDevice>> get_neighbours()
         const final;
+    virtual std::shared_ptr<sim::ILink> next_inlink() final;
     virtual std::shared_ptr<sim::ILink> get_link_to_destination(
-        std::shared_ptr<IRoutingDevice> device) const final;
+        std::shared_ptr<IRoutingDevice> dest) const final;
+    virtual std::set<std::shared_ptr<sim::ILink>> get_outlinks() const final;
```
PaulRalnikov, line 28: Please, remove all `virtual`

test/simulator/recalculate_paths_test.cpp
```diff
@@ -0,0 +1,359 @@
+#include <gtest/gtest.h>
```
PaulRalnikov, line 1: Let's remove gigantic number of lines line `EXPECT_EQ(first_device->get_link_to_destination(first_device)->get_to(second_device), third_device);` to testing effect:
Create function `bool check_reachibility(cource_device, destination_device)` that checks is it possible to reach `destination_device` from `source_device` going throught links from routing_table.

In test run this function for all pairs of devices you use -> profit!

test/simulator/utils.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include "simulator.hpp"
+
+namespace test {
+
+void add_links(
+    sim::Simulator& s,
+    std::initializer_list<std::pair<std::shared_ptr<sim::IRoutingDevice>,
+                                    std::shared_ptr<sim::IRoutingDevice>>>
+        links) {
```
PaulRalnikov, line 11: Use `using`:
```suggestion
using two_way_link_t = std::pair<std::shared_ptr<sim::IRoutingDevice>,
                                    std::shared_ptr<sim::IRoutingDevice> >;
void add_links(
    sim::Simulator& s,
    std::initializer_list<two_way_link_t>
        links) {
```

source/simulator.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (receiver== nullptr) {
+        LOG_WARN("add_flow failed: receiver is nullptr");
+        return;
+    }
+    m_flows.emplace_back(std::make_shared<Flow>(sender, receiver, 0.f));
+}
+
+void Simulator::add_link(std::shared_ptr<IRoutingDevice> a_from,
+                         std::shared_ptr<IRoutingDevice> a_to,
+                         std::uint32_t a_speed_mbps, std::uint32_t a_delay) {
```
ArtyomPeshkov, line 56: Use Time instead of std::uint32_t
```suggestion
                         std::uint32_t a_speed_mbps, Time a_delay) {
```

source/simulator.cpp
```diff
@@ -0,0 +1,126 @@
...
+            }
+            bool status = src_device->update_routing_table(
+                dest_device, parent_table[next_hop]);
+            LOG_INFO("Update routing table {} --> {}: {}", src_device_name,
+                         dest_device_name, (status ? "true" : "false"));
+        }
+    }
+}
+
+void Simulator::start(std::uint32_t a_stop_time) {
```
ArtyomPeshkov, line 116: Same here:
```suggestion
void Simulator::start(Time a_stop_time) {
```

test/simulator/recalculate_paths_test.cpp
```diff
@@ -0,0 +1,359 @@
...
+namespace test {
+
+class RecalculatePaths : public testing::Test {
+public:
+    void TearDown() override {};
+    void SetUp() override {};
+};
+
+constexpr std::uint32_t speed_stub = 0;
+constexpr std::uint32_t delay_stub = 0;
```
ArtyomPeshkov, line 17: Time here:
```suggestion
constexpr Time delay_stub = 0;
```

test/switch/receiver_mock.hpp
```diff
@@ -10,12 +10,15 @@ class ReceiverMock : public sim::IReceiver {
...
-    std::vector<std::shared_ptr<IRoutingDevice>> get_neighbours() const final;
+    std::vector<std::shared_ptr<sim::IRoutingDevice>> get_neighbours()
+        const final;
     std::shared_ptr<sim::ILink> next_inlink() final;
     std::shared_ptr<sim::ILink> get_link_to_destination(
-        std::shared_ptr<IRoutingDevice> device) const final;
+        std::shared_ptr<IRoutingDevice> dest) const final;
+    std::set<std::shared_ptr<sim::ILink>> get_outlinks() const final;
 
     std::uint32_t process() final;
```
ArtyomPeshkov, line 23:
```suggestion
    Time process() final;
```

source/receiver.cpp
```diff
@@ -8,6 +8,7 @@
 #include "link.hpp"
 #include "scheduler.hpp"
 
+
```
ArtyomPeshkov, line 11: Double endl
```suggestion

```

source/receiver.cpp
```diff
@@ -22,12 +23,24 @@ bool Receiver::add_inlink(std::shared_ptr<ILink> link) {
         LOG_WARN("Link destination device is incorrect (expected current device)");
         return false;
     }
-    m_router->add_inlink(link);
-    return true;
+    return m_router->add_outlink(link);
```
ArtyomPeshkov, line 26: I think it ment to be inlink here:
```suggestion
    return m_router->add_inlink(link);

```

test/device/add_link_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+            next_outlink++;
+            if (next_outlink == outlinks.end()) {
+                next_outlink = outlinks.begin();
+            }
+        }
+    }
+
+    for (auto appearances : number_of_appearances) {
+        EXPECT_EQ(appearances.second, NUMBER_OF_LOOPS);
+    }
```
ArtyomPeshkov, line 108: I dont really like this copypaste here, It would be cool if you will extract the common part

source/receiver.cpp
AntoxaBarin: Of course, my bad

test/device/add_link_test.cpp
AntoxaBarin: Seems like the biggest problem here is outlinks iterator (inlinks have `next_inlink()` method, outlinks not). In general, do we need an iterator for outlinks?

@PaulRalnikov, what do you think?

source/device.cpp
```diff
@@ -57,4 +50,16 @@ std::shared_ptr<ILink> RoutingModule::next_inlink() {
     return link;
 }
 
+std::vector<std::shared_ptr<ILink>> RoutingModule::get_outlinks() const {
+    return m_outlinks;
+}
+
+std::vector<std::shared_ptr<IRoutingDevice>> RoutingModule::get_neighbours()
```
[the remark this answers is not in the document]
AntoxaBarin: I think it is the result of a bad merge. Now order corresponds to the `main` branch

source/device.hpp
AntoxaBarin: Yes, it was an unnecessary methods reordering

test/device/add_link_test.cpp
PaulRalnikov: You may implements interator logic in this file accordind to fact we dont need `next_outlink` as device method now. Please, do it and merge similar tests (`InLinkIsPresent`, with `OutLinkIsPresent`, `SameInlinkMultipleTimes` with `SameOutlinkMultipleTimes`)

test/simulator/utils.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include "device.hpp"
+#include "simulator.hpp"
+
+namespace test {
+
+using two_way_link_t = std::pair<std::shared_ptr<sim::IRoutingDevice>,
+                                 std::shared_ptr<sim::IRoutingDevice>>;
+
+void add_links(sim::Simulator& sim,
+               std::initializer_list<two_way_link_t> links);
```
PaulRalnikov, line 16: Lets use name that speaks for itself better:
```suggestion
void add_two_way_links(sim::Simulator& sim,
               std::initializer_list<two_way_link_t> links);
```

test/simulator/utils.cpp
```diff
@@ -0,0 +1,30 @@
...
+        src_device->get_link_to_destination(dest_device)->get_to();
+    while (curr_device != dest_device) {
+        if (curr_device == nullptr) {
+            return false;
+        }
+        curr_device =
+            curr_device->get_link_to_destination(dest_device)->get_to();
+    }
+    return true;
+}
```
PaulRalnikov, line 28: Incorrect routing table may contain loops that will lead to infinite `while`. To deal with it, mark devices you meet. Also, please, be careful with nullptr
```suggestion
static std::shared_ptr<sim::IRoutingDevice> > get_next_device(std::shared_ptr<sim::IRoutingDevice> curr_device,
                                                                                                     std::shared_ptr<sim::IRoutingDevice> dest_device) {
    auto next_link = curr_device->get_link_to_destination(dest_device);
    if (link == nullptr) {
        return nullptr;
    }
    return link->get_to();
}

bool check_reachability(std::shared_ptr<sim::IRoutingDevice> src_device,
                       std::shared_ptr<sim::IRoutingDevice> dest_device) {
    std::set<std::shared_ptr<sim::IRoutingDevice> > used;
    auto curr_device = src_device;
    while (curr_device != dest_device) {
        if (curr_device == nullptr || used.contains(curr_device) ) {
            return false;
        }
        used.insert(curr_device);
        curr_device = get_next_device(curr_device, dest_device);
    }
    return true;
}
```

test/simulator/recalculate_paths_test.cpp
```diff
@@ -0,0 +1,193 @@
...
+    auto sender1 = sim.add_device("sender1", sim::DeviceType::SENDER);
+    auto sender2 = sim.add_device("sender2", sim::DeviceType::SENDER);
+    auto swtch = sim.add_device("switch", sim::DeviceType::SWITCH);
+    auto receiver = sim.add_device("receiver", sim::DeviceType::RECEIVER);
+
+    add_links(sim, {{sender1, swtch}, {sender2, swtch}, {swtch, receiver}});
+
+    sim.recalculate_paths();
+
+    auto devices = {sender1, sender2, swtch, receiver};
```
PaulRalnikov, line 74: Lets add type for `devices`:
```suggestion
    auto devices = std::vector<std::shared_ptr<IRoutingDevice> > ({sender1, sender2, swtch, receiver}};
```

test/simulator/recalculate_paths_test.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    auto sender = sim.add_device("sender", sim::DeviceType::SENDER);
+    auto swtch = sim.add_device("switch", sim::DeviceType::SWITCH);
+    auto receiver = sim.add_device("receiver", sim::DeviceType::RECEIVER);
+
+    add_links(sim, {{sender, swtch}, {swtch, receiver}});
+
+    sim.recalculate_paths();
+
+    auto devices = {sender, swtch, receiver};
```
PaulRalnikov, line 39:
```suggestion
    auto devices =;
    auto devices = std::vector<std::shared_ptr<IRoutingDevice> > ({sender, swtch, receiver});
```

test/simulator/recalculate_paths_test.cpp
```diff
@@ -0,0 +1,193 @@
...
+                    {swtch1, receiver1},
+                    {swtch1, receiver2},
+                    {swtch1, swtch2},
+                    {swtch2, receiver2},
+                    {swtch2, receiver3}});
+
+    sim.recalculate_paths();
+
+    auto devices = {sender1, sender2,   sender3,   swtch1,
+                    swtch2,  receiver1, receiver2, receiver3};
```
PaulRalnikov, line 125: Same
```suggestion
    auto devices = std::vector<std::shared_ptr<IRoutingDevice> > ({sender1, sender2,   sender3,   swtch1,
                    swtch2,  receiver1, receiver2, receiver3});
```

PaulRalnikov:
Please, add some tests for `Simulator::start` to check full work of system

source/simulator.cpp
```diff
@@ -52,8 +52,10 @@ void Simulator::add_flow(std::shared_ptr<IRoutingDevice> a_from,
         LOG_WARN("add_flow failed: receiver is nullptr");
         return;
     }
-    m_flows.emplace_back(std::make_shared<Flow>(sender, receiver, packet_size,
-                                                delay_between_packets));
+    m_flows.emplace_back(
+        std::make_shared<Flow>(std::dynamic_pointer_cast<Sender>(a_from),
+                               std::dynamic_pointer_cast<Receiver>(a_to),
+                               packet_size, delay_between_packets));
```
AntoxaBarin, line 58: I hope it is temporary solution. In the future, when new types of devices appear, it is possible to implement template function like this:

```cpp
template <typename S, typename R>
requires std::derived_from<S, ISender> && std::derived_from<R, IReceiver>

void Simulator::add_flow(std::shared_ptr<IRoutingDevice> a_from,
                         std::shared_ptr<IRoutingDevice> a_to,
                         Time delay_between_packets, Size packet_size) {
    m_flows.emplace_back(
        std::make_shared<Flow>(std::dynamic_pointer_cast<S>(a_from),
                               std::dynamic_pointer_cast<R>(a_to),
                               packet_size, delay_between_packets));
}
```

PaulRalnikov: I think in future it is appropriate to make our simulator template, like this:
```
template<typename TSender, TReceiver, TSwitch, TFlow>
class Simulator {
    ...
}
```

test/simulator/recalculate_paths_test.cpp
```diff
@@ -0,0 +1,193 @@
...
+    auto devices = std::vector<std::shared_ptr<sim::IRoutingDevice>>{sender, swtch, receiver};
+
+    for (auto src : devices) {
+        for (auto dest : devices) {
+            if (src != dest) {
+                EXPECT_TRUE(check_reachability(src, dest));
+            }
+        }
+    }
+}
```
PaulRalnikov, line 48: Please, move common part of tests (two nested `for`) into static function

test/simulator/start_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+    constexpr Size packet_size = 1024;
+
+    sim.add_flow(sender, receiver, packet_size, delay_between_packets);
+
+    add_two_way_links(sim, {{sender, swtch}, {swtch, receiver}});
+    sim.recalculate_paths();
+
+    sim.start(stop_time);
+    
+    // Test something...
```
PaulRalnikov, line 31: Make you test meaningfull (test that alfter end of `sim.start()` paket sended by flow really arrived to receiver, last one formed ask packet and this packer arrived to sender). Because of that you need to implement some simple flow logic; (like send only one packet and do not control congestion)

test/simulator/start_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+    auto receiver = sim.add_device("receiver", sim::DeviceType::RECEIVER);
+
+    constexpr Time delay_between_packets = 100;
+    constexpr Time stop_time = 1000;
+    constexpr Size packet_size = 1024;
+
+    sim.add_flow(sender, receiver, packet_size, delay_between_packets);
+
+    add_two_way_links(sim, {{sender, swtch}, {swtch, receiver}});
+    sim.recalculate_paths();
```
PaulRalnikov, line 27: You should not run `recalculate_path`, simulator have to do it by yourself (and already do it now)
```suggestion
```

test/device/add_link_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+namespace test {
+
+class AddLink : public testing::Test {
+public:
+    void TearDown() override {};
+    void SetUp() override {};
+};
+
+std::shared_ptr<sim::ILink> next_outlink(
```
PaulRalnikov, line 17:
```suggestion
static std::shared_ptr<sim::ILink> next_outlink(
```

test/device/add_link_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+    std::shared_ptr<sim::IRoutingDevice> device) {
+    static auto outlinks = device->get_outlinks();
+    static auto next_outlink_it = outlinks.begin();
```
PaulRalnikov, line 20: Huh? `outlinks` ans `next_outlink_it` will initialized once, and passing multiple devices to this function may lead to unexpected behavior)/ I think it's time to create common code for this part and `RoutingDevice`. Please, create new template class `LoopIterator` that folds all logic (its constructor will have two pointers: `std::forward_iterator<T> begin` and `std::forward_iterator<T> end`). It will also have have `operator++` and `operator*`). Put it inside new direcory `source/utils`

PaulRalnikov: In `operator*` i think you may throw exception if `begin = end`

test/simulator/start_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+    sim::Simulator sim;
+    auto sender = sim.add_device("sender", sim::DeviceType::SENDER);
+    auto swtch = sim.add_device("switch", sim::DeviceType::SWITCH);
+    auto receiver = sim.add_device("receiver", sim::DeviceType::RECEIVER);
+
+    constexpr Time delay_between_packets = 100;
+    constexpr Time stop_time = 1000;
+    constexpr Size packet_size = 1024;
+
+    sim.add_flow(sender, receiver, packet_size, delay_between_packets);
```
PaulRalnikov, line 24: Typo:
```suggestion
    sim.add_flow(sender, receiver, delay_between_packets, packet_size);
```

source/simulator.cpp
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+}
+
+void Simulator::start(Time a_stop_time) {
+    recalculate_paths();
+    m_scheduler.add(std::make_unique<Stop>(a_stop_time));
+    for (std::shared_ptr<IFlow> flow : m_flows) {
+        flow->start(0);
+    }
```
PaulRalnikov, line 116: Please, add `Process` event for each device and `SendData` event for each `Sender`

source/simulator.hpp
```diff
@@ -26,8 +32,9 @@ class Simulator {
 
 private:
     Scheduler& m_scheduler;
```
PaulRalnikov, line 34: Remove Scheduler from the simulator please to avoid misleading. The Scheduler is singleton and used in other places as `Scheduler::get_instance()`
```suggestion
```

source/simulator.cpp
PaulRalnikov: Implemented in [fix simulator PR](https://github.com/Cloud-Storage-Team/algnet/pull/127)

source/simulator.hpp
PaulRalnikov: Implemented in https://github.com/Cloud-Storage-Team/algnet/pull/127

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,45 @@
...
+        if (begin == end) {
+            throw std::runtime_error("LoopIterator range cannot be empty");
+        }
+    }
+
+    value_ref_t operator*() const { return *m_curr; }
+
+    LoopIterator& operator++() {
+        ++m_curr;
+        if (m_curr == m_end) {
```
ArtyomPeshkov, line 27: Maybe like this?
```suggestion
        if (++m_curr == m_end) {
```

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,45 @@
...
+    using value_ref_t = typename std::iterator_traits<Iter>::reference;
+
+    LoopIterator() = default;
+
+    LoopIterator(Iter begin, Iter end)
+        : m_curr(begin), m_begin(begin), m_end(end) {
+        if (begin == end) {
+            throw std::runtime_error("LoopIterator range cannot be empty");
+        }
+    }
```
ArtyomPeshkov, line 21: We decided to use prefix a_ for arguments passed to constructors
```suggestion
    LoopIterator(Iter a_begin, Iter a_end)
        : m_curr(a_begin), m_begin(a_begin), m_end(a_end) {
        if (a_begin == a_end) {
            throw std::runtime_error("LoopIterator range cannot be empty");
        }
    }
```

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,44 @@
...
+    using value_ref_t = typename std::iterator_traits<Iter>::reference;
+
+    LoopIterator() = default;
+
+    LoopIterator(Iter a_begin, Iter a_end)
+        : m_curr(a_begin), m_begin(a_begin), m_end(a_end) {
+        if (a_begin == a_end) {
+            throw std::runtime_error("LoopIterator range cannot be empty");
+        }
+    }
```
PaulRalnikov, line 21: Let's change logic the extreme case `m_begin = m_end`. Let allow to construct such iterator, but in `operator++` of it just white error message to log and return `m_end`, in `operator*` write critical message to log and return dereferenced (and potentially invalid) value
```suggestion
    LoopIterator(Iter a_begin, Iter a_end)
        : m_curr(a_begin), m_begin(a_begin), m_end(a_end) {}
```

PaulRalnikov: > Let's change logic the extreme case `m_begin = m_end`. Let allow to construct such iterator, but in `operator++` of it just white error message to log and return `m_end`, in `operator*` write critical message to log and return dereferenced (and potentially invalid) value
> 
> To simply use iterator, in `next_[in\out]link` instead of check if the link container is empty, check that corresponding iterator is equal to the end of container

In the beginning of problem discussion i wrote you to throw and error in incorrect iterator. But later in discussion with @ArtyomPeshkov we found better solution (suggested now). Sorry for this vagueness and spending your time

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+    value_ref_t operator*() const { return *m_curr; }
+
+    LoopIterator& operator++() {
+        if (++m_curr == m_end) {
+            m_curr = m_begin;
+        }
+        return *this;
+    }
```
ArtyomPeshkov, line 30:
```suggestion
    LoopIterator& operator++() {
        if (m_begin == m_end) {
            LOG_ERROR("Loop's begin iterator equals to end iterator");
        }
        
        if (++m_curr == m_end) {
            m_curr = m_begin;
        }
        return *this;
    }
```

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+    LoopIterator operator++(int) {
+        auto iter = *this;
+        ++(*this);
+        return iter;
+    }
```
ArtyomPeshkov, line 36:
```suggestion
    LoopIterator operator++(int) {
        if (m_begin == m_end) {
            LOG_ERROR("Loop's begin iterator equals to end iterator");
        }
        
        auto iter = *this;
        ++(*this);
        return iter;
    }
```

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,44 @@
...
+    LoopIterator() = default;
+
+    LoopIterator(Iter a_begin, Iter a_end)
+        : m_curr(a_begin), m_begin(a_begin), m_end(a_end) {
+        if (a_begin == a_end) {
+            throw std::runtime_error("LoopIterator range cannot be empty");
+        }
+    }
+
+    value_ref_t operator*() const { return *m_curr; }
```
ArtyomPeshkov, line 23:
```suggestion
    value_ref_t operator*() const { 
        if (m_begin == m_end) {
            LOG_CRITICAL("Loop's begin iterator equals to end iterator while dereferencing");
        }
        
        return *m_curr; 
    }
```

ArtyomPeshkov (commented):
Changes for LoopIterator class

source/flow.hpp
```diff
@@ -12,7 +12,7 @@ class ISender;
 class IFlow {
 public:
     virtual void start(Time time) = 0;
-    virtual bool try_to_generate(Size packet_size) = 0;
+    virtual std::optional<Time> try_to_generate() = 0;
```
PaulRalnikov, line 15: Rollback this interface change. it optimises time of simulator work but occures many unneseccary changes. Moreover, not is is not essential optimisation.
```suggestion
    virtual Time try_to_generate() = 0;
```

source/flow.hpp
```diff
@@ -23,29 +23,41 @@ class IFlow {
...
+    std::uint32_t get_updates_number() const;
 
     std::shared_ptr<ISender> get_sender() const final;
     std::shared_ptr<IReceiver> get_receiver() const final;
 
+protected:
+    void schedule_packet_generation(Time time);
+    void generate_packet();
+
 private:
```
PaulRalnikov, line 51: Make it private:
```suggestion
private:
    void schedule_packet_generation(Time time);
    void generate_packet();

```

source/flow.hpp
```diff
@@ -23,29 +23,41 @@ class IFlow {
...
-    ISender *m_src;
-    IReceiver *m_dest;
+    std::shared_ptr<ISender> m_src;
+    std::shared_ptr<IReceiver> m_dest;
     std::uint32_t m_nacks;
     float m_cwnd;
     Size m_sent_bytes;
```
PaulRalnikov, line 56: Please, remove redundant fields:
```suggestion
```

source/simulator.cpp
```diff
@@ -0,0 +1,167 @@
...
+    switch (a_type) {
+        case DeviceType::SENDER:
+            return add_sender(name);
+        case DeviceType::SWITCH:
+            return add_switch(name);
+        case DeviceType::RECEIVER:
+            return add_receiver(name);
+    }
+    return nullptr;
+}
```
PaulRalnikov, line 59: Lets delete this method from `Simulator` and will use only `add_receiver`, `add_sender` and `add_switch`
```suggestion
```

source/simulator.cpp
```diff
@@ -0,0 +1,167 @@
...
+                         std::shared_ptr<IRoutingDevice> a_to,
+                         std::uint32_t a_speed_mbps, Time a_delay) {
+    auto link = std::make_shared<Link>(a_from, a_to, a_speed_mbps, a_delay);
+    m_links.emplace_back(link);
+    a_from->add_outlink(link);
+    a_to->add_inlink(link);
+}
+
+// returns start device routing table
+routing_table_t bfs(std::shared_ptr<IRoutingDevice>& start_device) {
```
PaulRalnikov, line 82:
```suggestion
static routing_table_t bfs(std::shared_ptr<IRoutingDevice>& start_device) {
```

source/simulator.hpp
```diff
@@ -1,22 +1,39 @@
...
-    IRoutingDevice* add_device(std::string a_name, DeviceType a_type);
-    void add_flow(ISender* a_from, IReceiver* a_to);
-    void add_link(IRoutingDevice* a_from, IRoutingDevice* a_to,
+    Simulator() = default;
+    ~Simulator() = default;
+    std::shared_ptr<Sender> add_sender(std::string name);
+    std::shared_ptr<Receiver> add_receiver(std::string name);
+    std::shared_ptr<Switch> add_switch(std::string name);
+    std::shared_ptr<IRoutingDevice> add_device(std::string a_name,
+                                               DeviceType a_type);
```
PaulRalnikov, line 23:
```suggestion
```
See my comment below

test/simulator/start_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+                              delay_between_packets, packets_to_send_by_flow1);
+    auto flow2 = sim.add_flow(sender2, receiver, packet_size,
+                              delay_between_packets, packets_to_send_by_flow2);
+    auto flow3 = sim.add_flow(sender3, receiver, packet_size,
+                              delay_between_packets, packets_to_send_by_flow3);
+
+    add_two_way_links(sim, {{sender1, swtch},
+                            {sender2, swtch},
+                            {sender3, swtch},
+                            {swtch, receiver}});
```
PaulRalnikov, line 60: Change code parts order for better readbility
```suggestion
    add_two_way_links(sim, {{sender1, swtch},
                            {sender2, swtch},
                            {sender3, swtch},
                            {swtch, receiver}});
                            
    constexpr Time delay_between_packets = 100;
    constexpr Time stop_time = 10000;
    constexpr Size packet_size = 1024;
    constexpr std::uint32_t packets_to_send_by_flow1 = 10;
    constexpr std::uint32_t packets_to_send_by_flow2 = 50;
    constexpr std::uint32_t packets_to_send_by_flow3 = 100;

    auto flow1 = sim.add_flow(sender1, receiver, packet_size,
                              delay_between_packets, packets_to_send_by_flow1);
    auto flow2 = sim.add_flow(sender2, receiver, packet_size,
                              delay_between_packets, packets_to_send_by_flow2);
    auto flow3 = sim.add_flow(sender3, receiver, packet_size,
                              delay_between_packets, packets_to_send_by_flow3);
```

test/simulator/start_test.cpp
```diff
@@ -0,0 +1,106 @@
...
```
PaulRalnikov, line 94: Same as below
```suggestion
    add_two_way_links(sim, {{sender1, swtch},
                            {sender2, swtch},
                            {sender3, swtch},
                            {swtch, receiver}});

    constexpr Time delay_between_packets = 100;
    constexpr Time stop_time = 1000;
    constexpr Size packet_size = 1024;
    constexpr std::uint32_t packets_to_send_by_flow1 = 10;
    constexpr std::uint32_t packets_to_send_by_flow2 = 50;
    constexpr std::uint32_t packets_to_send_by_flow3 = 100;

    auto flow1 = sim.add_flow(sender1, receiver, packet_size,
                              delay_between_packets, packets_to_send_by_flow1);
    auto flow2 = sim.add_flow(sender2, receiver, packet_size,
                              delay_between_packets, packets_to_send_by_flow2);
    auto flow3 = sim.add_flow(sender3, receiver, packet_size,
                              delay_between_packets, packets_to_send_by_flow3);

```

AntoxaBarin:
After commit c4dedfb it takes 5 minutes instead of 2 to run `test` CI job

AntoxaBarin:
> After commit [c4dedfb](https://github.com/Cloud-Storage-Team/algnet/commit/c4dedfbd1bfa3a7adbb92c7814593e1b0b059b07) it takes 5 minutes instead of 2 to run `test` CI job

With commented logger in `source/utils/LoopIterator.hpp` CI job runs normal 2 minutes

source/utils/loop_iterator.hpp
```diff
@@ -0,0 +1,63 @@
+#pragma once
+
+#include <iterator>
+
+namespace sim {
+
+struct LOG {
+    static void CRITICAL(std::string_view msg);
+    static void ERROR(std::string_view msg);
+};
```
AntoxaBarin, line 10: Now LoopIterator calls static methods of struct LOG, logger.hpp includes into loop_iterator.cpp, `test` job runs in 2 minutes as normal.

PaulRalnikov:
> After commit [c4dedfb](https://github.com/Cloud-Storage-Team/algnet/commit/c4dedfbd1bfa3a7adbb92c7814593e1b0b059b07) it takes 5 minutes instead of 2 to run `test` CI job

I see common 2 minutes duration of CI after this commit.

source/utils/loop_iterator.hpp
PaulRalnikov, line 10: Please, remove this redundant interlayer
```suggestion
```

PaulRalnikov:
Thanks, good work! Will be approved after resolving comments

PaulRalnikov (approved):
Thanks! Let's merge it and make release!

PR #69, opened 2025-02-14
Network Simulator: Device

ArtyomPeshkov:


source/device.hpp
```diff
@@ -17,6 +17,12 @@ class Device {
     virtual ~Device();
     virtual void process();
 
+    void add_inlink(Link* link);
+    void add_outlink(Link* link, Device* dest);
```
drgulevich, line 21: Done in https://github.com/Cloud-Storage-Team/algnet/pull/49

source/device.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+namespace sim {
+
+void Device::add_inlink(Link* link) {
+    m_inlinks.insert(link);
+
+    // TODO: maybe some update of m_next_inlink is needed here
+}
+
+void Device::add_outlink(Link* link, Device* dest) {
```
drgulevich, line 11: This is now called `update_routing_table`. But as soon as we have `update_routing_table`, I am not convinced that `add_outlink` is needed at all.

ArtyomPeshkov:
Do I need to change branch from api to main?

AntoxaBarin:
Should we add getters to containers with links?  Without them simulator can't destroy links in the end of simulation and traverse net graph in `recalculate_path()`.

drgulevich:
> Do I need to change branch from api to main?

Yes, but preferably in a way that does not inherit the `api` branch commit history.

PaulRalnikov (approved):

drgulevich:
The PR description should say which issue it closes.

PR #70, opened 2025-02-14
FEAT: Flow

### LINKED ISSUE #50: Network Simulator: implementation of the Flow class methods

BogruAKVD:
Closes #50


BogruAKVD:
Как я вижу поддержку Flow со стороны Event'ов:
* Делается start(), где создается Event - Generate
* Generate в operator(): вызывает Flow.try_to_generate(). Если false то перездаём ивент с другим временем и кладём в Scheduler.

Поддержка Flow со стороны Sender'а:
* Вызывает Flow.ack(), если пришёл ACK пакет для этого Flow
* Вызываает Flow.trigger_congestion(), если перегрузка по этому Flow

drgulevich:
> Как я вижу поддержку Flow со стороны Event'ов:
> 
>     * Делается start(), где создается Event - Generate
> 
>     * Generate в operator(): вызывает Flow.try_to_generate(). Если false то перездаём ивент с другим временем и кладём в Scheduler.
> 
> 
> Поддержка Flow со стороны Sender'а:
> 
>     * Вызывает Flow.ack(), если пришёл ACK пакет для этого Flow
> 
>     * Вызывает Flow.trigger_congestion(), если перегрузка по этому Flow

We agreed to use English for GitHub communications.

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    bool Flow::try_to_generate(size_t current_time) { 
+        if (m_packets_in_flight < m_cwnd && m_packets_sent<m_total_packets) { //maybe good to stop generating events if all packets are sent
+            schedule_packet_generation(std::max(m_last_send_time + m_delay_between_packets, current_time));
+            return true;
+        }
+        return false;
+    }
+
+    void Flow::update(size_t delay){
```
drgulevich, line 35: Which congestion control algorithm is this?

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
...
+        if (delay < 100){
```
drgulevich, line 36: What is a magic number 100?

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
...
+            ack();
+        } else {
+            trigger_congestion();
+        }
+    }
+
+    void Flow::ack() {
```
drgulevich, line 43: Why separating this into `ack` and `trigger_congestion`?

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
...
+        m_ssthresh = m_cwnd / 2; 
+        m_cwnd = 1;
+    }
+
+    void Flow::schedule_packet_generation(int time) {
+        Generate generate_event(this, m_packet_size);
+        generate_event.time = time;
+        Scheduler::get_instance().add(generate_event);
+        ++m_packets_in_flight;
+        ++m_packets_sent;
```
drgulevich, line 66: Why `++m_packets_sent` not `m_packets_sent++`?

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
+#include "flow.hpp"
+#include "event.hpp"
+#include "scheduler.hpp"
+#include "packet.hpp"
+#include "sender.hpp"
+#include <iostream>
+
+namespace sim {
+
+    Flow::Flow(Device* src, size_t packet_size, size_t delay_between_packets, size_t total_packets)
```
drgulevich, line 10: Why do we need `total_packets`?

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
...
+        : m_src(src), 
+          m_packet_size(packet_size), 
+          m_delay_between_packets(delay_between_packets),
```
drgulevich, line 13: Do we need `delay_between_packets`?

source/flow.cpp
```diff
@@ -0,0 +1,70 @@
...
+          m_total_packets(total_packets),
+          m_packets_sent(0),
+          m_cwnd(1), 
+          m_ssthresh(128), 
+          m_linear_coeficient(1),
+          m_packets_in_flight(0), 
+          m_packets_acked(0),
+          m_last_send_time(0) {}
+
+    void Flow::start(size_t time){
```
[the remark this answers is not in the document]
BogruAKVD: I thought (maybe confused with the comment in the review) that we changed the simulator's start time to size_t (it seems logical that it cannot be negative) and decided to do the same with Flow

source/flow.cpp
BogruAKVD: This has also not been discussed, but in my opinion it is a good way to simulate threads with different loads. If you need to send packets as often as possible, you can make it 0. By the way, with zero latency, it's good to limit the number of packets.

source/flow.cpp
BogruAKVD: Of course, this has not been discussed, but the need to send a limited amount of data seems quite normal.

source/flow.cpp
BogruAKVD: Replaced it with the parameter

source/flow.cpp
drgulevich: The conversation should not be marked "resolved" without response to the comment. Also, it is better to leave it to the reviewer to confirm that this has been resolved indeed.

source/flow.cpp
drgulevich: Please, respond to the comment.

source/flow.cpp
drgulevich: ?

source/flow.cpp
drgulevich: There is not need to make it more complicated than it should be. Our aim is to make the code as simple as possible.

source/flow.cpp
drgulevich: No need to make it complex. Please, think in terms how we can make the code simple, not how we can make it complex. The timing between packet dispatches will be defined by the NIC rate on the sender.

drgulevich:
I proposed to introduce a base class Flow, from which other types of flows are derived (according to the specific algorithms).

drgulevich:
No PR description: which issue does it close?

source/flow.cpp
BogruAKVD: prefix increment can be slightly faster. But i rewrite to use postfix

source/flow.cpp
BogruAKVD: Made a comment in the code, but will be virtual (without implementing a specific algorithm).

source/flow.cpp
BogruAKVD: Rewrote it to a single update method

source/flow.hpp
```diff
@@ -3,31 +3,62 @@
...
+    void schedule_packet_generation(std::uint32_t time);
+
+    std::uint32_t m_packet_size;
+    std::uint32_t m_delay_between_packets;
+    std::uint32_t m_total_packets;
+    std::uint32_t m_packets_sent;
+    std::uint32_t m_last_send_time;
+
+    std::uint32_t m_cwnd;      // Congestion window
+    std::uint32_t m_ssthresh;  // Slow start threshold
```
drgulevich, line 57: As discussed last week, could we, please, have:
1. Have a single base class `Flow` in a single PR.
2. Other classes with inherit from the base class with the specific algorithms on separate PRs.

BogruAKVD:
> No PR description: which issue does it close?



> No PR description: which issue does it close?

drgulevich (commented):
Thanks, see comments attached

PaulRalnikov:
Please, do `git merge main`

source/flow.hpp
```diff
@@ -6,28 +6,45 @@
...
+
+    virtual ~IFlow() = default;
+};
+
+class Flow : public IFlow {  // Fixed inheritance
+public:
+    Flow(ISender* a_src, IReceiver* a_dest, uint32_t a_packet_size);
+    void start(std::uint32_t time) override = 0;
+    bool try_to_generate(std::uint32_t packet_size) override = 0;
+    void update(std::uint32_t delay) override = 0;
```
ArtyomPeshkov, line 33: This functions are not virtual, so they cant be = 0
```suggestion
    void start(std::uint32_t time) override;
    bool try_to_generate(std::uint32_t packet_size) override;
    void update(std::uint32_t delay) override;
```

ArtyomPeshkov, line 33: override and = 0 cant be used at the same time
```suggestion
    void start(std::uint32_t time) override;
    bool try_to_generate(std::uint32_t packet_size) override;
    void update(std::uint32_t delay) override;
```

BogruAKVD: This is necessary in order not to implement these methods in Flow

PR #80, opened 2025-02-21
FEAT: Receiver

### LINKED ISSUE #62: Network Simulator: implementation of the Receiver class methods

ArtyomPeshkov:
Closes #62


source/receiver.hpp
```diff
@@ -14,7 +14,7 @@ class Receiver : public Device {
     // Upon receiving send an ACK to the sender.
     // Packets are taken from ingress buffers on a round-robin basis.
     // The iterator over ingress buffers is stored in m_next_link.
-    void process();
+    void process(std::uint32_t start_time);
```
PaulRalnikov, line 17: Why do you need `start_time`? Now it used only for computation of time of next `Process` event. I think, it should be done in  `Process`'es `operator()`

PR #91, opened 2025-02-28
Network Simulator: Sender

ArtyomPeshkov:


.gitignore
```diff
@@ -1,4 +1,39 @@
+<<<<<<< HEAD
```
drgulevich, line 1: Resolution of conflicts is needed

PR #93, opened 2025-02-28
REFACTOR: Device interface isolation

### LINKED ISSUE #92: Simulator project: Device interface isolation
- Split `Device` interface into two: `IProcessingDevice` and `IRoutingDevice`.
- Rename `Device` into `RoutingModule` and make it heir of `IRoutingDevice`
- Create interfaces for classes `Sender`, `Receiver` and `Switch` and use composition instead of inherinace in them
- Replace raw pointers by smart ones in all mentioned classes and interfaces

PaulRalnikov:
Closes #92

### DIFF
--- a/source/device.cpp
+++ b/source/device.cpp
@@ -8,9 +8,8 @@ void Device::add_inlink(Link* link) {
     // TODO: maybe some update of m_next_inlink is needed here
 }
 
-void Device::update_routing_table(Device* dest, Link* link) {
+void Device::update_routing_table(IDevice* dest, Link* link) {
     m_routing_table[dest] = link;
 }
-
 
 }  // namespace sim
--- a/source/device.cpp
+++ b/source/device.cpp
@@ -2,17 +2,17 @@
 
 namespace sim {
 
-void RountingModule::add_inlink(std::shared_ptr<Link> link) {
+void RoutingModule::add_inlink(std::shared_ptr<Link> link) {
     m_inlinks.insert(link);
     m_next_inlink = m_inlinks.begin();
 }
 
-void RountingModule::update_routing_table(std::shared_ptr<IRoutingDevice> dest,
-                                          std::shared_ptr<Link> link) {
+void RoutingModule::update_routing_table(std::shared_ptr<IRoutingDevice> dest,
+                                         std::shared_ptr<Link> link) {
     m_routing_table[dest] = link;
 }
 
-std::shared_ptr<Link> RountingModule::next_inlink() {
+std::shared_ptr<Link> RoutingModule::next_inlink() {
     if (m_inlinks.empty()) {
         return nullptr;
     }
--- a/source/device.cpp
+++ b/source/device.cpp
@@ -2,14 +2,25 @@
 
 namespace sim {
 
-void Device::add_inlink(Link* link) {
+void RountingModule::add_inlink(Link* link) {
     m_inlinks.insert(link);
-
-    // TODO: maybe some update of m_next_inlink is needed here
+    m_next_inlink = m_inlinks.begin();
 }
 
-void Device::update_routing_table(IDevice* dest, Link* link) {
+void RountingModule::update_routing_table(IRoutingDevice* dest, Link* link) {
     m_routing_table[dest] = link;
+}
+
+Link* RountingModule::next_inlink() {
+    if (m_inlinks.empty()) {
+        return nullptr;
+    }
+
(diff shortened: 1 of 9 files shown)


source/flow.hpp
```diff
@@ -22,8 +23,8 @@ class Flow {
     void update();
 
 private:
-    Device *m_src;
-    Device *m_dest;
+    IRoutingDevice *m_src;
+    IRoutingDevice *m_dest;
```
ArtyomPeshkov, line 27: It seems like these interfaces were meant here:
```suggestion
    ISender *m_src;
    IReceiver *m_dest;
```

PaulRalnikov: Yes, I are right

source/simulator.hpp
```diff
@@ -13,10 +13,10 @@ class Simulator {
 public:
     Simulator();
     ~Simulator();
-    Device* add_device(std::string a_name, DeviceType a_type);
-    void add_flow(Device* a_from, Device* a_to);
-    void add_link(Device* a_from, Device* a_to, std::uint32_t a_speed_mbps,
-                  std::uint32_t a_delay);
+    IRoutingDevice* add_device(std::string a_name, DeviceType a_type);
+    void add_flow(IRoutingDevice* a_from, IRoutingDevice* a_to);
```
ArtyomPeshkov, line 17: And the same thing here:
```suggestion
    void add_flow(ISender* a_from, IReceiver* a_to);
```

PaulRalnikov: Yes, my miss

source/device.cpp
```diff
@@ -2,15 +2,26 @@
 
 namespace sim {
 
-void Device::add_inlink(Link* link) {
+void RountingModule::add_inlink(std::shared_ptr<Link> link) {
```
drgulevich, line 5: Naming: RoutingModule (without 'n')

source/device.hpp
```diff
@@ -9,25 +10,49 @@ class Link;
 
 enum DeviceType { SWITCH, SENDER, RECEIVER };
 
-class Device {
+class IProcessableDevice {
```
drgulevich, line 13: The intention was to say that this is a device that processes something. In this case, this should be `IProcessingDevice`. The word "processable" has a passive meaning: a device which could be processed, which is not what we mean here.

source/event.hpp
```diff
@@ -44,8 +44,8 @@ struct Arrive : public Event {
  * and start processing at the device.
  */
 struct Process : public Event {
-    Process(Device *a_device);
-    Device *node;
+    Process(RountingModule *a_device);
```
drgulevich, line 47: Should `RoutingModule` be `RoutingDevice`? In this case, 
```
Process(RoutingDevice *a_device);
```
would look better

drgulevich (commented):
Thanks, nice work. Could, you, please articulate here why we need such splitting (for the convenience of testing). From the first sight it looks like the both classes are used simultaneously as a composition in the inheritance and the need for such splitting is not very obvious.

source/device.cpp
PaulRalnikov: Thanks, missspell fixed

source/device.hpp
PaulRalnikov: Ok, I agree with you

source/event.hpp
PaulRalnikov: I think it would be better to abstract away here:
```suggestion
    Process(IProcessingDevice *a_device);
```

PaulRalnikov:
> Thanks, nice work. Could, you, please articulate here why we need such splitting (for the convenience of testing). From the first sight it looks like the both classes are used simultaneously as a composition in the inheritance and the need for such splitting is not very obvious.

We need to have interface for each class to be able to mock them. Also, we want to have some common in device classes part for routing. Because of that looks logical to create class for routing part - `RoutingModule`. There are to ways to use it: composition or inheritance. Code with inheritance looks like:
`class IReceiver : public IRoutingDevice, IProcessingDevice {
public:
    virtual ~IReceiver() = default;
};

class Receiver : public IReceiver, RoutingModule`
I think it is not clear that `RoutingModule` implements come methods from `IReceiver` interface so a see this idea not really good. Using compositions we gets much more clear for understanding code

drgulevich (approved):
Thanks! LGTM

ArtyomPeshkov (approved):

PR #99, opened 2025-03-07
FEAT: Logger

### LINKED ISSUE #97: FEAT: add logger
Add some info logs about actions, that happen on devices
Create logger configuration

AntoxaBarin:
Closes #97 

Add [spdlog](https://github.com/gabime/spdlog) logger as git submodule. It is a header-only version of spdlog.

[Examples of usage](https://github.com/gabime/spdlog/blob/v1.x/example/example.cpp).

To add a submodule you just need to run the following command:
`git submodule update --init --recursive`.

Then build project as usual:
```bash
mkdir build && cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
cmake --build .
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "googletest"]
 	path = googletest
 	url = https://github.com/google/googletest.git
+[submodule "spdlog"]
+	path = spdlog
+	url = https://github.com/gabime/spdlog.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,8 +7,8 @@ project(simulator)
 
 option(BUILD_PROJECT "Build tests" ON)
 
+include_directories(${CMAKE_SOURCE_DIR}/spdlog/include/)
 if (BUILD_PROJECT)
-    include_directories(${CMAKE_SOURCE_DIR}/spdlog/include/)
     set(TARGET_NAME "${PROJECT_NAME}")
     file(GLOB_RECURSE src "source/*.cpp")
     add_executable(${TARGET_NAME} main.cpp ${src})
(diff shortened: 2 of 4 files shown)


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "googletest"]
 	path = googletest
 	url = https://github.com/google/googletest.git
+[submodule "spdlog"]
+	path = spdlog
+	url = https://github.com/gabime/spdlog.git
```
drgulevich, line 6: Is this added automatically when adding a submodule?

drgulevich (commented):
Thanks! Is this the header-only or compiled version? Could you, please, provide a more detailed installation instruction to make this work.

.gitmodules
AntoxaBarin: Yes.

AntoxaBarin: It is a header-only version of the library. Added [installation instruction](https://github.com/Cloud-Storage-Team/algnet/pull/99#issue-2902780905).

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ project(simulator)
 option(BUILD_PROJECT "Build tests" ON)
 
 if (BUILD_PROJECT)
+    include_directories(${CMAKE_SOURCE_DIR}/spdlog/include/)
```
PaulRalnikov, line 11: I think it would be better to include our logger always:
```suggestion
include_directories(${CMAKE_SOURCE_DIR}/spdlog/include/)
if (BUILD_PROJECT)
```

AntoxaBarin: I agree with you.

PaulRalnikov (approved):

PR #117, opened 2025-04-03
FEAT: create type for Time

ArtyomPeshkov:


source/simulator.hpp
```diff
@@ -16,13 +16,13 @@ class Simulator {
...
     void add_link(IRoutingDevice* a_from, IRoutingDevice* a_to,
-                  std::uint32_t a_speed_mbps, std::uint32_t a_delay);
+                  std::uint32_t a_speed_mbps, Time a_delay);
     // Clear all events in the Scheduler
     void clear();
     // BFS to update the routing table
     void recalculate_paths();
     // Create a Stop event at a_stop_time and start simulation
-    void start(std::uint32_t a_stop_time);
+    void start(Time a_stop_time);
```
AntoxaBarin, line 25: What is 1 Time? A second? I think it is better to add several time types, for example `Time::ms`, `Time::ns`, `Time::s` and methods to convert values between them (`Time::s::to_ms()`, etc.).

This is just my point of view, correct me if I'm wrong.

I want to set simulation time as simple as possible:

```cpp
sim::Simulator S;
...
S.start(Time::s(10));
```

ArtyomPeshkov: In our system, most events happen in a few nanoseconds, so it seems to me that 1 Time should be 1 nanosecond.

source/types.hpp
```diff
@@ -0,0 +1,4 @@
+#include <cstdint>
+
+typedef std::uint32_t Time;
```
PaulRalnikov, line 3: According to [conversation](https://github.com/Cloud-Storage-Team/algnet/pull/117#discussion_r2027908916), I think such comment is appropriate
```suggestion
// nanoseconds
typedef std::uint32_t Time;
```

PaulRalnikov (approved):
Thanks! LGTM

AntoxaBarin (approved):

PR #154, opened 2025-04-24
FIX: compile Logger class

### LINKED ISSUE #153: BUG: compilation error for logger.cpp on Windows
`logger.cpp` does not compile on Windows machine complaining about the `Logger::ERROR` method ("expression can not be used as a function").

drgulevich:
Closes #153


test/switch/test_switch.cpp
```diff
@@ -1,14 +1,15 @@
 #include <gtest/gtest.h>
 
+#include <algorithm>
```
drgulevich, line 3: `std::sort` is not found without this header on Windows

PR #182, opened 2025-05-22
REFACTORING: Store time in scheduler

ArtyomPeshkov:
1) Now it's possible to get and update time of current event
2) Fixed bug with incorrect link arrival time calculation
3) Fixed bug with 0 number of packets in flows by default

### DIFF
--- a/source/link.cpp
+++ b/source/link.cpp
@@ -42,10 +42,8 @@ void Link::schedule_arrival(Packet packet) {
     LOG_INFO("Packet arrived to link's ingress queue. Packet: " + packet.to_string());
 
     unsigned int transmission_time = get_transmission_time(packet);
-    unsigned int total_delay = m_src_egress_delay + transmission_time;
-    (void) total_delay; // unused variable stub
-
     m_src_egress_delay += transmission_time;
+    unsigned int total_delay = Scheduler::get_instance().get_current_time() + m_src_egress_delay;
 
     // TODO: put correct event time. Arrive happens in current time + total_delay.
     Scheduler::get_instance().add(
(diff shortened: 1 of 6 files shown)


source/scheduler.cpp
```diff
@@ -15,6 +15,7 @@ bool Scheduler::tick() {
         std::move(const_cast<std::unique_ptr<Event>&>(m_events.top()));
     m_events.pop();
     event->operator()();
+    m_current_event_local_time = event->get_time();
```
PaulRalnikov, line 18: I think we should swap this two lines
```suggestion
    m_current_event_local_time = event->get_time();
    event->operator()();
```

source/link.cpp
```diff
@@ -9,27 +9,33 @@
...
 
 Time Link::get_transmission_time(const Packet& packet) const {
-    if (m_speed_mbps == 0) {
+    if (m_speed_gbps == 0) {
         LOG_WARN("Passed zero link speed");
         return 0;
     }
-    return (packet.size + m_speed_mbps - 1) / m_speed_mbps +
+    const std::uint32_t byte_to_bit_multiplier = 8;
+    const std::uint64_t gbit_to_bit_multiplier = 1024 * 1024 * 1024;
```
AntoxaBarin, line 33: 1 gigabit to bit = 1'000'000'000 bit
```suggestion
    const std::uint64_t gbit_to_bit_multiplier = 1'000'000'000;
```

source/parser.cpp
```diff
@@ -45,10 +45,10 @@ uint32_t YamlParser::parse_throughput(const std::string& throughput) {
     const uint32_t value = std::stoul(throughput.substr(0, unit_pos));
     const std::string unit = throughput.substr(unit_pos);
     if (unit == "Gbps") {
-        return value * 1000;  // Convert to Mbps
+        return value;
     }
     if (unit == "Mbps") {
-        return value;
+        return value / 1024;  // Convert to Gbps
```
AntoxaBarin, line 51:
```suggestion
        return value / 1000;  // Convert to Gbps
```

PaulRalnikov (approved):
LGTM!

AntoxaBarin (approved):

PR #195, opened 2025-05-25
Removed put_data_to_device from flow interface

PaulRalnikov:
Deleted `put_data_to_device` from `IFlow` interface. Reason: we have method `create_data_packet` that creates packet that will be send in time choosen by flow, so there is no need to make method like `put_data_to_device` public


ArtyomPeshkov (approved):
LGTM!

PR #222, opened 2025-05-29
Add TcpAimdFlow

AntoxaBarin:
Contributes to #268.

### DIFF
--- a/source/flow/tcp_aimd_flow.cpp
+++ b/source/flow/tcp_aimd_flow.cpp
@@ -32,9 +32,12 @@ TcpAimdFlow::TcpAimdFlow(Id a_id, std::shared_ptr<ISender> a_src,
 }
 
 void TcpAimdFlow::start() {
-    Generate generate_event(Scheduler::get_instance().get_current_time(),
-                            shared_from_this(), m_packet_size);
+    Time curr_time = Scheduler::get_instance().get_current_time();
+    Generate generate_event(curr_time, shared_from_this(), m_packet_size);
     Scheduler::get_instance().add(std::move(generate_event));
+
+    TcpMetric metrics_event(curr_time, shared_from_this());
+    Scheduler::get_instance().add(std::move(metrics_event));
 }
 
 Time TcpAimdFlow::create_new_data_packet() {
@@ -92,6 +95,8 @@ std::shared_ptr<IReceiver> TcpAimdFlow::get_receiver() const {
     return m_dest.lock();
 }
 Id TcpAimdFlow::get_id() const { return m_id; }
+
+std::uint32_t TcpAimdFlow::get_cwnd() const { return m_cwnd; }
 
 std::string TcpAimdFlow::to_string() const {
     std::ostringstream oss;
--- a/source/flow/tcp_aimd_flow.cpp
+++ b/source/flow/tcp_aimd_flow.cpp
@@ -5,7 +5,9 @@
 #include <algorithm>
 #include <sstream>
 
+#include "device/sender.hpp"
 #include "event.hpp"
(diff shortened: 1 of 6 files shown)


source/simulator.hpp
```diff
@@ -105,7 +106,8 @@ class Simulator {
         }
     }
     // Create a Stop event at a_stop_time and start simulation
-    void start(Time a_stop_time, bool export_metrics = false, bool draw_plots = false) {
+    void start(Time a_stop_time, bool export_metrics = false,
+               bool draw_plots = false) {
```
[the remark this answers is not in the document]
AntoxaBarin: Maybe remove `a_` prefix from all arguments?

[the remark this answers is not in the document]
PaulRalnikov: Added

source/parser/identifiable_parser.hpp
```diff
@@ -192,4 +194,29 @@ Id parse_object<TcpFlow>(const YAML::Node& key_node,
     return id;
 }
 
+template <>
+Id parse_object<TcpAimdFlow>(const YAML::Node& key_node,
```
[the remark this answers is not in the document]
PaulRalnikov: Issue #239 created

source/simulator.hpp
```diff
@@ -96,20 +97,23 @@ class Simulator {
...
-                for (auto [link, paths_count]: links) {
-                    src_device->update_routing_table(dest_device_id, link.lock(), paths_count);
+                for (auto [link, paths_count] : links) {
+                    src_device->update_routing_table(dest_device_id,
+                                                     link.lock(), paths_count);
                 }
             }
         }
     }
     // Create a Stop event at a_stop_time and start simulation
```
drgulevich, line 107: Nit: As `a_stop_time` has been changed to `stop_time`, the comment should be amended accordingly.

source/flow/tcp_aimd_flow.cpp
```diff
@@ -0,0 +1,135 @@
...
+        std::make_unique<TcpMetric>(curr_time, shared_from_this());
+    Scheduler::get_instance().add(std::move(metrics_event));
+}
+
+Time TcpAimdFlow::create_new_data_packet() {
+    if (m_packets_to_send == 0) {
+        return 0;
+    }
+    if (try_to_put_data_to_device()) {
+        --m_packets_to_send;
```
drgulevich, line 53: Nit: imho, post-decrement `m_packets_to_send--` looks better when they are equivalent

source/flow/tcp_aimd_flow.cpp
```diff
@@ -0,0 +1,135 @@
...
+        }
+        m_packets_acked++;
+
+        // Additive Increase
+        m_cwnd += 1;
+        try_to_put_data_to_device();
+    } else {
+        // Multiplicative Decrease
+        m_ssthresh = std::max(1u, static_cast<std::uint32_t>(m_cwnd * m_beta));
+        m_cwnd = m_ssthresh;
```
drgulevich, line 89: We are not using `m_ssthresh` in this algorithm. Can we get rid of `m_ssthresh` then?

source/flow/tcp_aimd_flow.cpp
```diff
@@ -0,0 +1,135 @@
...
+        m_packets_in_flight = 0;
```
drgulevich, line 90: What does this mean? How can we change `m_packets_in_flight`? The number of packets in flight is not something that we should control ourselves as it is just information about the number of packets flying in the network. It should only be changed only by either sending a packet (`m_packets_in_flight++`) or receiving and ACK ( `m_packets_in_flight--`).

PR #249, opened 2025-06-04
REFACTOR: Link

drgulevich:
Minor refactoring for the Link.

### DIFF
--- a/source/link.cpp
+++ b/source/link.cpp
@@ -59,18 +59,15 @@ void Link::schedule_arrival(Packet packet) {
         return;
     }
 
-    LOG_INFO("Packet is assigned to the link. Packet: " + packet.to_string());
-
     unsigned int transmission_time = get_transmission_time(packet);
     m_last_src_egress_pass_time =
         std::max(m_last_src_egress_pass_time,
                  Scheduler::get_instance().get_current_time()) +
         transmission_time;
 
-    Size old_value = m_src_egress_buffer_size_byte;
     m_src_egress_buffer_size_byte += packet.size_byte;
     MetricsCollector::get_instance().add_queue_size(
-        get_id(), Scheduler::get_instance().get_current_time(), old_value,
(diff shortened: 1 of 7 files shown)


source/link.cpp
```diff
@@ -23,8 +23,8 @@ Link::Link(Id a_id, std::weak_ptr<IRoutingDevice> a_from,
       m_last_src_egress_pass_time(0),
       m_transmission_delay(a_delay),
       m_next_ingress(),
-      m_ingress_buffer_size_byte(0),
-      m_max_ingress_buffer_size_byte(a_max_ingress_buffer_size) {
+      m_next_ingress_buffer_size_byte(0),
+      m_max_next_ingress_buffer_size_byte(a_max_ingress_buffer_size) {
```
drgulevich, line 27: Although this becomes more verbose, `...next_ingress...` makes it more clear that this refers to the next device ingress buffer.

PaulRalnikov: Maybe its better to use `...next_device_ingress...`?

source/link.cpp
```diff
@@ -54,58 +54,49 @@ void Link::schedule_arrival(Packet packet) {
...
 };
 
 void Link::process_arrival(Packet packet) {
-    if (m_ingress_buffer_size_byte + packet.size_byte >
-        m_max_ingress_buffer_size_byte) {
-        LOG_ERROR("Ingress buffer on Link overflow; packet" +
-                  packet.to_string() + " lost");
+    if (m_next_ingress_buffer_size_byte + packet.size_byte >
+        m_max_next_ingress_buffer_size_byte) {
+        LOG_ERROR("Ingress buffer overflow; packet" + packet.to_string() +
```
PaulRalnikov, line 80:
```suggestion
        LOG_ERROR("Ingress buffer overflow; packet " + packet.to_string() +
```

PR #252, opened 2025-06-04
REFACTOR: Change RTT to a proper *round* trip time

### LINKED ISSUE #250: Make RTT to be the *round* trip time
At the initial stages we have only taken into account the downstream time (from source to receiver) to make things simple. As we now have a proper simulation with ACKs being sent upstream, we are now ready to make RTT to be the actual *round* trip time as it should be.

drgulevich:
Closes #250. Packet class is refactored to notify of the sending time of the original data packet to calculate the full RTT.

### DIFF
--- a/source/device/receiver.cpp
+++ b/source/device/receiver.cpp
@@ -4,10 +4,10 @@
 
 #include "event.hpp"
 #include "link.hpp"
-#include "routing_module.hpp"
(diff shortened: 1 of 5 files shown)


drgulevich:
Ready for review.

drgulevich:
I approve myself and merge in order to move on. The changes here seem straightforward.

PR #258, opened 2025-06-05
FEAT: add ECN support

### LINKED ISSUE #175: Support for ECN

ArtyomPeshkov:
Closes #175.

### DIFF
--- a/source/device/switch.cpp
+++ b/source/device/switch.cpp
@@ -86,9 +86,9 @@ Time Switch::process() {
                     next_link->get_max_from_egress_buffer_size())) {
         packet.congestion_experienced = true;
     }
+    // TODO: increase total_processing_time correctly
     next_link->schedule_arrival(packet);
 
-    // TODO: increase total_processing_time correctly
     if (m_process_scheduler.notify_about_finish(
             Scheduler::get_instance().get_current_time() +
             total_processing_time)) {
--- a/source/device/switch.cpp
+++ b/source/device/switch.cpp
@@ -1,6 +1,7 @@
 #include "switch.hpp"
 
 #include <memory>
+#include <iostream>
 
 #include "link.hpp"
 #include "routing_module.hpp"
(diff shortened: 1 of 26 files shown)


source/device/switch.cpp
```diff
@@ -78,6 +82,10 @@ Time Switch::process() {
              packet.to_string());
 
     // TODO: increase total_processing_time correctly
+    double current_threshold = static_cast<double>(next_link->get_current_from_egress_buffer_size()) / static_cast<double>(next_link->get_max_from_egress_buffer_size());
```
drgulevich, line 85: Few comments here:
1. `current_threshold` is not a good naming because this is not a threshold here.
2. It is better to use multiplication to avoid division (more expensive operation) when possible. The condition to check is just:
```
static_cast<double>(next_link->get_max_from_egress_buffer_size()) >= m_ecn_threshold * static_cast<double>(next_link->get_current_from_egress_buffer_size());
```
which also looks a bit simpler.

drgulevich (commented):
Thanks for this! Just a quick look. I will look into it in more details a bit later.

source/packet.hpp
```diff
@@ -26,6 +26,8 @@ struct Packet {
     Size size_byte;
     IFlow* flow;
     Time send_time;
+    bool is_ecn_enabled;
+    bool detected_congestion;
```
drgulevich, line 30: Nit: naming suggestions:
```
    bool ecn_enabled;
    bool ecn_detected_congestion;
```
or
```
    bool ecn_capable_transport;
    bool congestion_experienced;
```
to match ECT and CE bit naming (https://datatracker.ietf.org/doc/html/rfc3168).

source/link.cpp
```diff
@@ -129,6 +129,11 @@ std::shared_ptr<IRoutingDevice> Link::get_to() const {
     return m_to.lock();
 };
 
+
+Size Link::get_current_from_egress_buffer_size() const {
```
drgulevich, line 133: The naming (motivated by `m_from_egress_buffer_size`) is confusing here. By buffer we actually mean *queue*. I suggest the following naming changes:

`get_current_from_egress_buffer_size` -> `get_current_from_egress_queue_size`

And we need to change `m_from_egress_buffer_size` to `m_from_egress_queue_size` (possibly in a separate refactoring PR).

drgulevich (commented):
Thanks. Few comments above.

source/link.cpp
drgulevich: I renamed  `m_from_egress_buffer_size` to `m_from_egress_queue_size` in https://github.com/Cloud-Storage-Team/algnet/pull/263

drgulevich:
I address the feedback and merge to move on.

PR #271, opened 2025-06-09
FEAT: Add the incast simulation example

### LINKED ISSUE #265: Add a configuration example for the incast scenario
This should be configurable via specifying the number of `N` identical senders, 1 switch and 1 receiver.

drgulevich:
Closes #265.


configuration_examples/generator/bottleneck.py
```diff
@@ -142,7 +163,7 @@ def main():
     print(f"Topology file saved as {args.topology} with {args.senders} senders and {args.receivers} receivers")
     
     # Generate simulation
-    simulation = generate_simulation(args.topology_path, args.senders, args.receivers)
+    simulation = generate_simulation(args.topology_path, args.senders, args.receivers, args.flows)
```
drgulevich, line 166: I have combined the two scripts `simple_...` and `cross_...` (they only differ by a couple of lines) into a single one with the `--flows` flag.

PR #274, opened 2025-06-10
FEAT: Delivery rate for each flow

### LINKED ISSUE #273: Add delivery rate
Add calculation of the delivery rate. A field which records the amount of data sent by the flow should be added to the packet and returned back to the origin with ACK (this is exactly how it is done in the BBR algorithm).

drgulevich:
Closes #273.  New field added to the packet which records the amount of data sent by the flow at the time of sending the packet (as it is done in BBR).

### DIFF
--- a/source/device/receiver.cpp
+++ b/source/device/receiver.cpp
@@ -102,7 +102,10 @@ Time Receiver::send_ack(Packet data_packet) {
     // Note: ACK's sent time is the data packet sent time
     Packet ack(PacketType::ACK, 1, data_packet.flow,
                data_packet.flow->get_receiver()->get_id(),
-               data_packet.flow->get_sender()->get_id(), data_packet.sent_time);
+               data_packet.flow->get_sender()->get_id(), data_packet.sent_time,
+               data_packet.sent_bytes_at_origin,
+               data_packet.ecn_capable_transport,
+               data_packet.congestion_experienced);
 
     std::shared_ptr<ILink> link_to_dest = get_link_to_destination(ack);
     if (link_to_dest == nullptr) {
--- a/source/device/switch.cpp
+++ b/source/device/switch.cpp
@@ -79,7 +79,8 @@ Time Switch::process() {
     LOG_INFO("Processing packet from link on switch. Packet: " +
              packet.to_string());
 
-    if (packet.ecn_capable_transport &&
+    // ECN mark for data packets
+    if (packet.type == DATA && packet.ecn_capable_transport &&
         static_cast<double>(next_link->get_from_egress_queue_size()) >=
             m_ecn_threshold *
                 static_cast<double>(
--- a/source/flow/tcp_flow.hpp
+++ b/source/flow/tcp_flow.hpp
@@ -28,6 +28,7 @@ private:
 
     bool try_to_put_data_to_device();
 
+    Id m_id;
     std::weak_ptr<ISender> m_src;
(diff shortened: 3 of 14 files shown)


source/device/switch.cpp
```diff
@@ -79,7 +79,8 @@ Time Switch::process() {
     LOG_INFO("Processing packet from link on switch. Packet: " +
              packet.to_string());
 
-    if (packet.ecn_capable_transport &&
+    // ECN mark for data packets
+    if (packet.type == DATA && packet.ecn_capable_transport &&
```
drgulevich, line 83: An addition to https://github.com/Cloud-Storage-Team/algnet/pull/258: ECN mark for DATA packets only

source/device/receiver.cpp
```diff
@@ -102,7 +102,10 @@ Time Receiver::send_ack(Packet data_packet) {
     // Note: ACK's sent time is the data packet sent time
     Packet ack(PacketType::ACK, 1, data_packet.flow,
                data_packet.flow->get_receiver()->get_id(),
-               data_packet.flow->get_sender()->get_id(), data_packet.sent_time);
+               data_packet.flow->get_sender()->get_id(), data_packet.sent_time,
+               data_packet.sent_bytes_at_origin,
+               data_packet.ecn_capable_transport,
+               data_packet.congestion_experienced);
```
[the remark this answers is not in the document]
drgulevich: In addition to https://github.com/Cloud-Storage-Team/algnet/pull/258: ACK sends the ECN info back

source/metrics/metrics_storage.cpp
```diff
@@ -47,6 +47,9 @@ matplot::figure_handle MetricsStorage::get_picture(
 
     ax->plot(x_data, y_data, "-o")->line_width(1.5);
 
+    auto limits = ax->ylim();
+    ax->ylim({0, limits[1]});
+
```
PaulRalnikov, line 52: Lets move this code from here to `draw_plot`. If some other function that uses `get_picture` adds some data to plot, it may not appear because of upper y limit. Now this bug appears in `MetricsCollector::draw_metrics_plots` in part about queue_size 
```suggestion
```

source/metrics/metrics_collector.cpp
```diff
@@ -92,6 +101,16 @@ void MetricsCollector::draw_metric_plots(
                                       flow->get_sender()->get_id(),
                                       flow->get_receiver()->get_id())});
     }
+
+    for (auto& [flow_id, values] : m_rate_storage) {
+        auto flow =
+            IdentifierFactory::get_instance().get_object<IFlow>(flow_id);
+        values.draw_plot(metrics_dir / fmt::format("rate/{}.svg", flow_id),
+                         {"Time, ns", "Values, packets",
```
PaulRalnikov, line 109:
```suggestion
                         {"Time, ns", "Values, Gbps",
```

source/flow/tcp_flow.hpp
```diff
@@ -41,6 +42,6 @@ class TcpFlow : public ITcpFlow, public std::enable_shared_from_this<TcpFlow> {
     std::uint32_t m_packets_in_flight;
     std::uint32_t m_packets_acked;
 
-    Id m_id;
```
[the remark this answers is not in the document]
PaulRalnikov: No, its bug.

drgulevich:
Thanks for the review! I merge this to move on.

PR #294, opened 2025-06-18
REFACTORING: scheduler

### LINKED ISSUE #290: REFACTORING: Scheduler
Change `Scheduler::add` to that one:
```
template<typename TEvent, typename Args...>
void Scheduler::add(Args&&... args) {
    static_assert(
        std::is_constructible_v<TEvent, Args&&...>,
        "Event must be constructible from given args"
    );
    static_assert(
        std::is_invocable_r<TEvent>::value,
        "Event must be functor"
    );
    
    //pure logic (creating unique_ptr, converting to std::variant etc)
    ...
}
```

PaulRalnikov:
closes #290


source/scheduler.hpp
```diff
@@ -24,7 +24,16 @@ class Scheduler {
         return instance;
     }
 
-    void add(std::unique_ptr<Event> event);
+    template <typename TEvent, typename... Args>
+    void add(Args&&... args) {
+        static_assert(std::is_constructible_v<TEvent, Args&&...>,
+                      "Event must be constructible from given args");
+        static_assert(std::is_base_of_v<Event, TEvent>,
+                      "TEvent must be inherit from Event");
```
AntoxaBarin, line 32: Typo
```suggestion
                      "TEvent must inherit from Event");
```

source/scheduler.hpp
```diff
@@ -24,7 +24,16 @@ class Scheduler {
...
+
+        m_events.push(std::make_unique<TEvent>(args...));
```
AntoxaBarin, line 34: Is `emplace` a better choice here?
```suggestion
        m_events.emplace(std::make_unique<TEvent>(args...));
```

PR #313, opened 2025-06-24
REFACTORING: Events and link

### LINKED ISSUE #312: REFACTORING: events & link
- Split `event.hpp` and `event.cpp` into many files: one pair .hpp and .cpp file for each event
- Make `Arrive` event accessible only inside `Link` class
- Delete `process_arrival` from `ILink` interface

PaulRalnikov:
closes #312


source/event/tcp_metric.cpp
```diff
@@ -0,0 +1,24 @@
+#include "flow/interfaces/i_tcp_flow.hpp"
```
drgulevich, line 1: I suggest this event to be deprecated (see [319](https://github.com/Cloud-Storage-Team/algnet/issues/319))

PaulRalnikov: Yes, it will be deleted soon. I will add task about it to the board

PR #322, opened 2025-06-26
BUG: Link transmission time

### LINKED ISSUE #276: BUG: Link transmission time
We currently have:
```
    unsigned int transmission_time = get_transmission_time(packet);
    m_arrival_time =
        std::max(m_arrival_time, Scheduler::get_instance().get_current_time()) +
        transmission_time;
```
I suspect that `transmission_time` here should be the bare transmission time without inclusion of the speed of light latency.

We have also discussed the possibility of introducing the proper queues as queue data structures.

PaulRalnikov:
closes #276


ArtyomPeshkov (approved):
LGTM!

PR #324, opened 2025-06-27
FIX: Improve stress topology generation scripts

### LINKED ISSUE #303: FIX: Fix stress topology generation scripts
1) Scripts should generate topology with N senders, N receivers and N flows in total going between i-th sender and i-th receiver, with 1 or 2 switches in between
2) Scripts should generate topology with N senders, M receivers and M * N flows in total (each to each), with 1 or 2 switches in between

AntoxaBarin:
Closes #303 

- Added README.md with examples of using a `bottleneck.py`
- Added several switches (between senders and receivers) support in `bottleneck.py`

### DIFF
--- a/configuration_examples/generator/README.md
+++ b/configuration_examples/generator/README.md
@@ -1,22 +1,36 @@
 # How to generate a config
 
-### Incast topology N-to-1
+All the following commands must be run from root directory!
 
-To generate simulation and topology configs for **N-to-1 incast** topology just run this command from root directory:
+### Incast topology N-to-1
 
 ```bash
 python3 configuration_examples/generator/bottleneck.py \
   --senders N \ # N should be replaced with a number
   --receivers 1 \
   --packets 10000 \
-  --simulation-time 1000000
-  --topology incast-50-to-1-topology.yml \
+  --simulation-time 1000000 \
+  --topology incast-N-to-1-topology.yml \
   --topology-dir configuration_examples/generator/ \
-  --simulation incast-50-to-1-simulation.yml \
+  --simulation incast-N-to-1-simulation.yml \
   --simulation-dir configuration_examples/generator/
 ```
 
-### Incast topology M-to-N
+### Incast topology N-to-N, N flows
+
+```bash
+python3 configuration_examples/generator/bottleneck.py \
+  --senders N \ # N should be replaced with a number
+  --receivers N \ # N should be replaced with a number
+  --packets 10000 \
+  --simulation-time 1000000 \
+  --topology incast-N-to-N-topology.yml \
+  --topology-dir configuration_examples/generator/ \
+  --simulation incast-N-to-N-simulation.yml \
+  --simulation-dir configuration_examples/generator/
+```
+
+### Incast topology M-to-N, M*N flows
 
 ```bash
 python3 configuration_examples/generator/bottleneck.py \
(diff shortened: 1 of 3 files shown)


configuration_examples/generator/README.md
```diff
@@ -0,0 +1,61 @@
+# How to generate a config using bottleneck.py
```
PaulRalnikov, line 1:
```suggestion
# Usage examples
```

configuration_examples/generator/bottleneck.py
```diff
@@ -3,11 +3,21 @@
...
+    num_switches,
+    switch_name="switch",
+    link_latency="0ns",
+    link_throughput="100Gbps",
+    ingress_buffer_size="1024000B",
+    egress_buffer_size="1024000B",
+):
+    topology = {"devices": {}, "links": {}}
+
+    sender_switch_name = "switch0"
```
PaulRalnikov, line 19: It will work incorrect if `swtich_name != switch`. I think best way to refactor it is to create array of switch names:
```
switch_names = [f"{swtich_name}_{i}" for i in range(num_swtiches)]
```

configuration_examples/generator/bottleneck.py
```diff
@@ -3,11 +3,21 @@
 import sys
 import os
 
-def generate_topology(num_senders, num_receivers, switch_name="switch", link_latency="0ns", link_throughput="100Gbps", ingress_buffer_size = "1024000B", egress_buffer_size = "1024000B"):
-    topology = {
-        "devices": {},
-        "links": {}
-    }
+
+def generate_topology(
```
PaulRalnikov, line 7: Lets unify device and links naming generatited by this function. I think the best way is to use `{device_name}_{device_num}` due to snake case alredy used for links naming (also its more readable)

configuration_examples/generator/bottleneck.py
```diff
@@ -46,43 +56,78 @@ def generate_topology(num_senders, num_receivers, switch_name="switch", link_lat
...
+            "ingress_buffer_size": ingress_buffer_size,
+            "egress_buffer_size": egress_buffer_size,
+        }
 
     return topology
 
-def generate_simulation(topology_file, num_senders, num_receivers, flows, simulation_time, packet_interval,
-                        number_of_packets, packet_size=1500, algorithm="tcp"):
+
+def generate_simulation(
```
PaulRalnikov, line 109: I can not add comment to not modified lines, so I write it here. I think its not good idea to use hardcoded device names (e.g. `f"sender{i}"` on 140 line, 'f"receiver{i}" in 141 line') because naming changes in `generate_topology` will have no affect to this function. I see two ways to deal with it:
1) Parse topology file in this function and get device names from it;
2) Pass topology configuration as dict to this function;

Second one looks better to me due to absence parsing self-generated text, but you may use any you like or suggest your one.

scripts/run_incast_simulation.py
```diff
@@ -64,14 +62,10 @@ def main(args):
...
-        "--topology-dir",
-        generator_dir_path,
-        "--simulation",
-        simulation_config_name,
-        "--simulation-dir",
-        generator_dir_path,
+        "--topology-path",
+        generator_dir_path + topology_config_name,
+        "--simulation-path",
+        generator_dir_path + simulation_config_name,
```
PaulRalnikov, line 68: Putting config files to `generator_dir_path` looks stange to me (this folder contains only scripts for configs generation). I think we may add new argument `--output-folder` which will containts both of configs. By default it might be set to `""` for simplicity

configuration_examples/generator/bottleneck.py
```diff
@@ -3,150 +3,234 @@
...
-    # Add the switch
-    topology["devices"][switch_name] = {"type": "switch", "threshold": 0.7}
+    # Add the switches
+    for i in range(0, num_switches):
+        topology["devices"][switch_names[i]] = {"type": "switch", "threshold": 0.7}
 
-    return topology
+    # Add links between switches
+    for i in range(0, num_switches - 1):
+        base_index = 2 * (num_senders + num_receivers) + 2 * i
```
PaulRalnikov, line 93: Lets change work with `base_index`: initialize it in the beginning of the function and use `base_index += 1` everywhere we need new id for link

PR #332, opened 2025-07-02
FEAT: add bitset for storing packet flags

ArtyomPeshkov:


source/flow/tcp_flow.cpp
```diff
@@ -151,4 +155,11 @@ bool TcpFlow::try_to_put_data_to_device() {
 
 double TcpFlow::get_cwnd() const { return m_cwnd; }
 
+void TcpFlow::initialize_flag_manager() {
+    if (!m_is_initialized) {
+        m_flags.register_flag_by_length("type", 1);
```
PaulRalnikov, line 160: Same here

source/utils/bitset.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "logger/logger.hpp"
+
+namespace sim {
+
+BitSet32::BitSet32() : m_data(0) {}
+
+BitSet32::BitSet32(std::uint32_t initial) : m_data(initial) {}
+
+bool BitSet32::set_bit(std::uint32_t pos, bool value) {
+    if (pos >= 32) {
```
PaulRalnikov, line 14: Please, use `sizeof(m_data)` instead of magic integer 32
```suggestion
    if (pos >= sizeof(m_data)) {
```

source/utils/bitset.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+    friend bool operator==(const BitSet32& fst, const BitSet32& snd);
+
+    void reset();
+
+private:
+    std::uint32_t m_data;
+
+    inline std::uint32_t max_range_value(std::uint32_t length) const {
+        return length == 32 ? 0xFFFFFFFF : (1U << length) - 1;
```
PaulRalnikov, line 28: You may use `std::numeric_limits`:
```suggestion
        return length == sizeof_bits(m_data) ? std::numeric_limits<std::uint32_t>::max(): (static_cast<std::uint32_t>1 << length) - 1;
```

source/utils/flag_manager.hpp
```diff
@@ -0,0 +1,93 @@
...
+            return 0;
+        }
+
+        std::uint32_t highest_bit = max_values - 1;
+        std::uint32_t bits_num = 0;
+        while (highest_bit) {
+            bits_num++;
+            highest_bit >>= 1;
+        }
+        return bits_num;
```
PaulRalnikov, line 86: You may use `__builtin_clz`:

```suggestion
        max_values--;
        return sizeof_bits(max_values) - __builtin_clz(max_values);
```

source/flow/tcp_flow.cpp
```diff
@@ -95,12 +98,13 @@ void TcpFlow::update(Packet packet, DeviceType type) {
...
     } else if (packet.dest_id == m_dest.lock()->get_id() &&
-               packet.type == PacketType::DATA) {
+               m_flags.get_flag(packet, "type") == PacketType::DATA) {
         // data packet delivered to destination device; send ack
-        Packet ack(PacketType::ACK, 1, this, m_dest.lock()->get_id(),
+        Packet ack(1, this, m_dest.lock()->get_id(),
                    m_src.lock()->get_id(), packet.sent_time,
                    packet.sent_bytes_at_origin, packet.ecn_capable_transport,
                    packet.congestion_experienced);
+        m_flags.set_flag(packet, "type", PacketType::ACK);
```
PaulRalnikov, line 107: Same
```suggestion
        if (!m_flags.set_flag(packet, "type", PacketType::ACK)) {
            throw std::runtime_error("Can not set ASK type to  packet");
        }
```

source/utils/bitset.hpp
```diff
@@ -0,0 +1,107 @@
...
+    explicit BitSet(BitStorage initial) : m_data(initial) {};
+
+    bool set_bit(std::uint8_t pos, bool value) {
+        if (pos >= sizeof_bits(m_data)) {
+            LOG_ERROR(fmt::format("Bit position is out of range. Max possible position is {}, but got {}", sizeof(BitStorage) - 1, pos));
+            return false;
+        }
+
+        if (value) {
+            m_data |= (1U << pos);
```
PaulRalnikov, line 37: Using `1U` might cause a bug with large types like `unsigned long long`. I think the best way to deal with this problem is to create `const static inline BitStorageType ONE = 1` and use it use it everywhere you need

source/utils/bitset.hpp
```diff
@@ -0,0 +1,107 @@
...
+    };
+
+    bool set_range(std::uint8_t low, std::uint8_t high, BitStorage value){
+        if (high >= sizeof_bits(m_data) || low > high) {
+            LOG_ERROR(fmt::format("Range edges error. Low and high should be less than {} and low should be <= hight. Got low = {} and high = {}", sizeof(BitStorage), low, high));
+            return false;
+        }
+
+        const std::uint32_t length = high - low + 1;
+        const std::uint32_t max_val = max_range_value(length);
```
PaulRalnikov, line 52: Typo/bug:

```suggestion
        const BitStorageType max_val = max_range_value(length);
```

source/utils/bitset.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+        if (value > max_val) {
+            LOG_ERROR(fmt::format("Value is out of range. Max possible value = {}, when got {}", max_val, value));
+            return false;
+        }
+
+        std::uint32_t mask = max_val << low;
```
PaulRalnikov, line 59: Same:

```suggestion
        BigStorageType mask = max_val << low;
```

source/utils/bitset.hpp
```diff
@@ -0,0 +1,107 @@
...
+        m_data = (m_data & ~mask) | (value << low);
+
+        return true;
+    };
+
+    std::uint8_t get_bit(std::uint8_t pos) const {
```
PaulRalnikov, line 65: Just a suggestion for better readability:

```suggestion
    book get_bit(std::uint8_t pos) const {
```

test/utils/bitset_test.cpp
```diff
@@ -0,0 +1,91 @@
+#include <gtest/gtest.h>
```
PaulRalnikov, line 1: Lets add some tests with types differs from `std::uint32_t`, like `unsigned long long` and `unsigned __int128`

source/flow/tcp_flow.hpp
```diff
@@ -25,10 +26,17 @@ class TcpFlow : public IFlow, public std::enable_shared_from_this<TcpFlow> {
...
+    static std::string packet_type_label;
+    enum PacketType { ACK, DATA, ENUM_SIZE };
+
     Packet generate_packet();
 
     bool try_to_put_data_to_device();
 
+    static void initialize_flag_manager();
+    static bool m_is_initialized;
+    static FlagManager<std::string, PacketFlagsBase> m_flags;
```
PaulRalnikov, line 38: Same here

```suggestion
    static bool m_is_flag_manager_initialized;
    static FlagManager<std::string, PacketFlagsBase> m_flag_manager;
```

source/packet.cpp
```diff
@@ -21,28 +21,14 @@ Packet::Packet(PacketType a_type, Size a_size_byte, IFlow* a_flow,
...
-        case PacketType::DATA:
-            oss << "DATA";
-            break;
-        default:
-            oss << "UNKNOWN";
-            break;
-    }
-
-    oss << ", source_id: " << source_id;
+    oss << "Packet[source_id: " << source_id;
```
PaulRalnikov, line 31: I think we should add flags to output too

source/utils/bitset.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+    BitStorage get_range(std::uint8_t low, std::uint8_t high) const {
+        if (high >= sizeof_bits(m_data)  || low > high) {
+            LOG_ERROR(fmt::format("Range edges error. Low and high should be less than {} and low should be <= hight. Got low = {} and high = {}", sizeof_bits(BitStorage), low, high));
+            return 0;
+        }
+
+        const std::uint8_t length = high - low + 1;
+        BitStorage mask = max_range_value(length) << low;
+        return (m_data & mask) >> low;
```
PaulRalnikov, line 78: Might be simplified:

```suggestion
        return (m_data >> low) & max_range_value(length);
```

source/utils/flag_manager.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+    bool register_flag_by_amount(FlagId id, BitStorage different_values) {
+        return register_flag_by_length(id, required_bits_for_values(different_values));
+    }
+
+    bool register_flag_by_length(FlagId id, BitStorage flag_length) {
+        if (flag_length == 0 || flag_length > sizeof_bits(BitStorage)) {
+            LOG_ERROR(fmt::format("Incorrect flag length. Max possible length is {}, flag length should be more than 0. Got {}", sizeof_bits(BitStorage), flag_length));
+            return false;
+        }
```
PaulRalnikov, line 26: Second condition here is redundant due to more stricter check in next `if`:

```suggestion
        if (flag_length == 0) {
            LOG_ERROR("Passed zero flag length, should be at least 1");
            return false;
        }
```

PR #358, opened 2025-07-14
FEAT: cwnd < 1 & pacing delay

### LINKED ISSUE #323: Support for CWND<1 by introducing pacing delay
CWND<1 is equivalent to introducing an artificial pacing delay (see, e.g. Swift paper https://dl.acm.org/doi/10.1145/3387514.3406591).

Let's fix the terminology as follows:
Pacing interval -- time interval between two packet dispatches: RTT/CWND
Pacing delay -- artificial delay introduced after ACK arrival (differs from pacing interval by RTT): RTT(1 - 1/CWND)

PaulRalnikov:
closes #323

### DIFF
--- a/source/flow/tcp/i_tcp_cc.hpp
+++ b/source/flow/tcp/i_tcp_cc.hpp
@@ -1,0 +1,14 @@
+#pragma once
+#include "types.hpp"
+
+namespace sim {
+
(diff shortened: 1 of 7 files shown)


ArtyomPeshkov (approved):
Thanks! LGTM!

PR #360, opened 2025-07-16
FEAT: specification of parse_object instead of parse_and_register

ArtyomPeshkov:


source/parser/identifiable_parser/identifiable_parser.hpp
```diff
@@ -10,13 +10,20 @@
...
+class IdentifieableParser  {
+    static_assert(std::is_base_of_v<Identifiable, T>);
+public:
+    // Parses object and return shared_ptr to it
+    static std::shared_ptr<T> parse_object(const YAML::Node& key_node,
+                                           const YAML::Node& value_node);
+};
+
+template <typename T>
+class IdentifieableRegister {
```
PaulRalnikov, line 22: I think `register` should be a private function and return `void`:
```
template <typename T>
class IdentifieableRegister {
static_assert(std::is_base_of_v<Identifiable, T>);

public:
    static std::shared_ptr<T> parse_and_registrate(
        const YAML::Node& key_node, const YAML::Node& value_node) {
        std::shared_ptr<T> object = IdentifieableParser<T>::parse_object(key_node, value_node);
        registrate(object);
        return object;
    }

private:
    static void registrate(std::shared_ptr<T> object) {
        if (!IdentifierFactory::get_instance().add_object(object)) {
            throw std::runtime_error(fmt::format(
                "Can not add object with type {}; object with same id ({}) "
                "already exists",
                typeid(T).name(), object.get()->get_id()));
        }
    }
}
```

source/parser/identifiable_parser/identifiable_parser.hpp
```diff
@@ -10,13 +10,20 @@
 namespace sim {
 
 template <typename T>
-class IdentifieableParser {
+class IdentifieableParser  {
```
PaulRalnikov, line 13: According to fact this class does not need `T` to be `Identifiable`, I think class name should be changed:
```suggestion
class Parser  {
```

source/parser/identifiable_parser/identifiable_parser.hpp
PaulRalnikov, line 22: Suggestion (might be done only if you accept [this one](https://github.com/Cloud-Storage-Team/algnet/pull/360/files#r2210278182):
```suggestion
class IdentifieableParser {
```

PR #370, opened 2025-07-20
FEAT: Add hasher parsing support

### LINKED ISSUE #373: FEAT: hasher parsing support
Add ability to set packet hasher in config

AntoxaBarin:
closes #373


source/parser/identifiable_parser/parse_host.cpp
```diff
@@ -1,13 +1,20 @@
...
-std::shared_ptr<Host> Parser<Host>::parse_object(
-    const YAML::Node& key_node, const YAML::Node& value_node) {
-    (void)value_node;
-    return std::make_shared<Host>(key_node.as<Id>());
+std::shared_ptr<Host> Parser<Host>::parse_object(const YAML::Node& key_node,
+                                                 const YAML::Node& value_node) {
+    std::unique_ptr<IHasher> hasher = nullptr;
+    if (value_node["hasher"]) {
+        hasher = parse_hasher(value_node);
+    }
```
PaulRalnikov, line 16: Please, use `parse_with_default`. Now it may work only with `uint32_t`, but its really easy to make it template. Please, do it

source/parser/identifiable_parser/parse_switch.cpp
```diff
@@ -13,10 +15,17 @@ template <>
 std::shared_ptr<Switch> Parser<Switch>::parse_object(
     const YAML::Node& key_node, const YAML::Node& value_node) {
     Id id = key_node.as<Id>();
+    std::unique_ptr<IHasher> hasher = nullptr;
+    if (value_node["hasher"]) {
+        hasher = parse_hasher(value_node);
+    }
```
PaulRalnikov, line 21: Same here

source/parser/parse_utils.cpp
```diff
@@ -52,4 +56,20 @@ uint32_t parse_with_default(
...
+        return std::make_unique<sim::RandomHasher>();
+    }
+    if (hasher_value == "base") {
+        return std::make_unique<sim::BaseHasher>();
+    }
+    if (hasher_value == "symmetric") {
+        return std::make_unique<sim::SymmetricHasher>();
+    }
+    LOG_WARN(fmt::format("Unknown hasher type: {}", hasher_value));
+    return std::make_unique<sim::BaseHasher>();
```
PaulRalnikov, line 74: I think its better to use `LOG_ERROR` and return `nullptr` here:
```suggestion
    LOG_ERROR(fmt::format("Unknown hasher type: {}", hasher_value));
    return nullptr;
```

ArtyomPeshkov: Do we really want to force user to specify hasher? I think it would be more convenient to just set some default value if hasher is not specified and write some warining log, as suggested [here](https://github.com/Cloud-Storage-Team/algnet/pull/370/files#r2222249119)

PaulRalnikov:
Closed until parsing refactoring

PR #378, opened 2025-07-24
BUG: delivery rate

### LINKED ISSUE #372: BUG: delivery rate
Now delivery rate of `TcpFlow` calculates incorrectly : now measures total size of packets that were **sent** per RTT. Need to replace it with total size of packets that were **acked** per RTT.

PaulRalnikov:
closes #372

### DIFF
--- a/configuration_examples/simulation_examples/basic_simulation.yml
+++ b/configuration_examples/simulation_examples/basic_simulation.yml
@@ -4,13 +4,13 @@ flows:
   flow1:
     sender_id: "sender1"
     receiver_id: "receiver1"
-    packet_size: 1024
-    number_of_packets: 10000
+    packet_size: 64
+    number_of_packets: 100
   flow2:
     sender_id: "sender2"
     receiver_id: "receiver2"
-    packet_size: 1024
-    number_of_packets: 10000
+    packet_size: 128
+    number_of_packets: 100
 
 algorithm: basic
 simulation_time: 100000
--- a/source/flow/tcp/tcp_flow.hpp
+++ b/source/flow/tcp/tcp_flow.hpp
@@ -28,7 +28,7 @@ public:
           m_packets_to_send(a_packets_to_send),
           m_ecn_capable(a_ecn_capable),
           m_packets_in_flight(0),
-          m_packets_acked(0),
+          m_delivered(0),
           m_sent_bytes(0),
           m_avg_rtt(0.0) {
         if (m_src.lock() == nullptr) {
@@ -48,7 +48,7 @@ public:
             m_flag_manager.get_flag(packet, packet_type_label) ==
                 PacketType::ACK) {
             // ACK delivered to source device; calculate metrics, update
-            // internal state
+            // internal state(ь_ - packet.delivered_at_origin) / rtt
             TimeNs current_time = Scheduler::get_instance().get_current_time();
             if (current_time < packet.sent_time) {
                 LOG_ERROR("Packet " + packet.to_string() +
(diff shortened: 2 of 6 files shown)


source/flow/tcp/tcp_flow.hpp
```diff
@@ -48,7 +48,7 @@ class TcpFlow : public IFlow,
             m_flag_manager.get_flag(packet, packet_type_label) ==
                 PacketType::ACK) {
             // ACK delivered to source device; calculate metrics, update
-            // internal state
+            // internal state(ь_ - packet.delivered_at_origin) / rtt
```
ArtyomPeshkov, line 51: If I understand this correctly, we dont need this one here
```suggestion
            // internal state
```

source/flow/tcp/tcp_flow.hpp
```diff
@@ -222,7 +221,7 @@ class TcpFlow : public IFlow,
     bool m_ecn_capable;
 
     std::uint32_t m_packets_in_flight;
-    std::uint32_t m_packets_acked;
+    SizeByte m_delivered;
```
ArtyomPeshkov, line 224: I think we should rename this to m_bytes_delivered, in the context of flow it's not clear if we speak about delivered packets or number of delivered bytes with current naming
```suggestion
    SizeByte m_delivered_bytes;
```

source/flow/tcp/tcp_flow.hpp
```diff
@@ -222,7 +221,7 @@ class TcpFlow : public IFlow,
...
     SizeByte m_sent_bytes;
```
ArtyomPeshkov, line 225: Maybe just remove this one now? It seems like we dont really use it anymore
```suggestion
```

configuration_examples/simulation_examples/basic_simulation.yml
```diff
@@ -4,13 +4,13 @@ flows:
...
-    number_of_packets: 10000
+    packet_size: 64
+    number_of_packets: 100
   flow2:
     sender_id: "sender2"
     receiver_id: "receiver2"
-    packet_size: 1024
-    number_of_packets: 10000
+    packet_size: 128
+    number_of_packets: 100
```
ArtyomPeshkov, line 13: Do you want to keep this changes or you just forgot to rollback them?

source/flow/tcp/tcp_flow.hpp
```diff
@@ -61,21 +61,21 @@ class TcpFlow : public IFlow,
...
-
             double old_cwnd = m_cc.get_cwnd();
 
             if (m_packets_in_flight > 0) {
                 m_packets_in_flight--;
             }
             if (!m_cc.on_ack(rtt, m_avg_rtt, packet.congestion_experienced)) {
                 // No congestion
-                m_packets_acked++;
+                m_delivered += m_packet_size;
```
ArtyomPeshkov, line 71: Ohhhh, I dont like this one, this will cause a lot of problems as soon as we start using better flows with packets of different size
I think we should use total number of transfered bytes per current flow as packet number and calculate amount of delivered bytes as difference of packet numbers (classic TCP-like stuff)

source/packet.hpp
```diff
@@ -26,7 +26,7 @@ struct Packet {
     IFlow* flow;
     TimeNs sent_time;  // Note: ACK's sent time is the data packet sent time
     SizeByte
-        sent_bytes_at_origin;  // For ACK this is inherited from data packet
+        delivered_at_origin;  // For ACK this is inherited from data packet
```
ArtyomPeshkov, line 29: And the same thing here
```suggestion
        delivered_bytes_at_origin;  // For ACK this is inherited from data packet
```

configuration_examples/simulation_examples/basic_simulation.yml
PaulRalnikov: Really want to keep

source/flow/tcp/tcp_flow.hpp
PaulRalnikov: Renamed to `m_delivered_data_size`

source/packet.hpp
PaulRalnikov: Renamed to `delivered_data_size_at_origin`

ArtyomPeshkov (approved):
LGTM! Thanks!

PR #411, opened 2025-08-06
REFACTOR: eliminate long double

drgulevich:
No need to use `long double` -- this is a huge overkill. Floating point type `double` is far than sufficient for our purpose (even `float` could be fine, but `double` is a gold standard for most purposes).


drgulevich:
Thanks for review

PR #412, opened 2025-08-06
FEAT: compile time logs disabling

### LINKED ISSUE #413: FEAT: compile time logs disabling
Add log level as compile option

PaulRalnikov:
closes #413


source/logger/logger.hpp
```diff
@@ -16,28 +16,68 @@ class Logger {
...
+    void critical(std::string&& msg, const std::source_location& =
+                                         std::source_location::current());
 
 private:
     Logger();
     Logger(const Logger&) = delete;
     Logger& operator=(const Logger&) = delete;
 };
 
+#define LOG_LEVEL_TRACE 0
```
AntoxaBarin, line 38: Can we use only `TRACE` without `LOG_LEVEL`?

AntoxaBarin: Also for other log levels macros

PR #420, opened 2025-08-15
FEAT: Fat-tree generator

### LINKED ISSUE #419: Three-level fat tree topology generation
We need to be able to generate 3-level fat tree topology parametrized by k -- number of ports per switch (see table in https://packetpushers.net/blog/demystifying-dcn-topologies-clos-fat-trees-part2/ for L=3). The most important fat tree cases include:

1. k=4: smallest possible 3-level fat tree (16 hosts).
2. k=16: minimal non-trivial 3-level fat tree used in a variety of research studies (1024 host). For example FASTFLOW (https://arxiv.org/abs/2404.01630) and other.

We might also need subsections of the 3-level fat tree like Fig.2a (single-rack only) and Fig.2b (leaf-spine only) in https://www.sciencedirect.com/science/article/abs/pii/S1389128624002895.

ArtyomPeshkov:
Closes #419.


scripts/topology_generator/fat_tree.py
```diff
@@ -0,0 +1,125 @@
...
+    config["links"][f"link{link_counter}"] = {
+        "from": from_node, "to": to_node, "preset-name": preset
+    }
+    link_counter += 1
+    config["links"][f"link{link_counter}"] = {
+        "from": to_node, "to": from_node, "preset-name": preset
+    }
+    return link_counter + 1
+
+def generate_fat_tree_config(k):
```
PaulRalnikov, line 13: I think name `k` is not really well
```suggestion
def generate_fat_tree_config(switch_ports_count):
```

scripts/topology_generator/fat_tree.py
```diff
@@ -0,0 +1,125 @@
...
+                    "ecn": {"min": 0.2, "max": 0.3, "probability": 1.0}
+                }
+            }
+        },
+        "packet-spraying": {"type": "ecmp"},
+        "hosts": {},
+        "switches": {},
+        "links": {}
+    }
+
```
PaulRalnikov, line 65: I think it would be nice to simplify hosts and switches naming. Also due to fact it used many times, creating lambdas for them makes sense
```suggestion
    host_name = labmda pod_idx, host_idx: f"pod{pod_idx}_host{host_idx}"
    arrg_name = lambda pod_idx, arrg_idx : f"pod{pod_idx}_arrg{aggr_idx}"
    edge_name = lambda pod_idx, edge_idx : f"pod{pod_idx}_arrg{edge_idx}"
```

scripts/topology_generator/fat_tree.py
```diff
@@ -0,0 +1,125 @@
...
+        "packet-spraying": {"type": "ecmp"},
+        "hosts": {},
+        "switches": {},
+        "links": {}
+    }
+    
+    for i in range(1, senders + 1):
+        config["hosts"][f"sender{i}"] = {"layer": 3}
+    for i in range(1, receivers + 1):
+        config["hosts"][f"receiver{i}"] = {"layer": 3}
```
PaulRalnikov, line 69: Why do you split host names into senders and receivers?
```suggestion
    for p in range(1, num_pods + 1):
        for h in range(1, hosts_per_pod + 1):
            config["hosts"][host_name(p, h)] = {"layer": 3}
```

scripts/topology_generator/fat_tree.py
```diff
@@ -0,0 +1,125 @@
...
+    
+    # Edge-host
+    for pod in range(num_pods):
+        for edge_idx in range(1, edge_per_pod + 1):
+            edge_id = pod * edge_per_pod + edge_idx
+            for h in range(1, k // 2 + 1):
+                host_num = pod * hosts_per_pod + (edge_idx - 1) * (k // 2) + h
+                host_type = "sender" if host_num <= total_hosts // 2 else "receiver"
+                actual_host_num = host_num if host_type == "sender" else host_num - senders
+                link_counter = add_bidirectional_link(config, f"{host_type}{actual_host_num}", f"edge{edge_id}", link_counter)
```
PaulRalnikov, line 92: Same. Also I think it would be nice to add new variable `hosts_per_edge` with value `k // 2` (used in suggestion)
```suggestion
    for pod_idx in range(1, num_pods + 1):
        for edge_idx in range(1, edge_per_pod + 1):
            for h in range(1, hosts_per_edge + 1):
                host_idx = (edge_idx - 1) * hosts_per_edge + h
                link_counter = add_bidirectional_link(config, host_name(pod_idx, host_idx), edge_name(pod_idx, edge_idx), link_counter)
```

scripts/topology_generator/fat_tree.py
```diff
@@ -0,0 +1,125 @@
...
+    
+    with open(filename, 'w') as f:
+        yaml.dump(config, f, sort_keys=False, width=120, indent=2)
+    
+    print(f"Configuration written to {filename}")
+    return filename
+
+if __name__ == "__main__":
+    k_value = 8
+    output_file = write_config_to_file(k_value)
```
PaulRalnikov, line 125: And I think we should add `-o` ('--output-file') option to script that describes path to generated file

PR #425, opened 2025-08-25
BUG: Fix delivery rate

drgulevich:
Delivery rate should take into account the pacing delay.


drgulevich:
Additional feature: added throughput per flow to summary.

PR #454, opened 2025-09-23
FEAT: Adding data sending scenarios

Az3git:


source/parser/simulation/scenario/action/send_data_parser.cpp
```diff
@@ -0,0 +1,101 @@
...
+            result.emplace_back(str.substr(start, end - start + 1));
+    }
+
+    return result;
+}
+
+// Get target connections from the YAML node
+static std::vector<std::weak_ptr<IConnection>> get_target_connections(
+    const YAML::Node& node) {
+    std::vector<std::weak_ptr<IConnection>> conns;
```
PaulRalnikov, line 29: We do not need any other syntax as regulars, `conn1,conn2,conn3` may be introduced by regular expression `conn[1-3]`

source/scenario/action/send_data_action.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace sim {
+
+class SendDataAction : public IAction {
+public:
+    SendDataAction(TimeNs a_when, SizeByte a_size,
+                   std::vector<std::weak_ptr<IConnection>> a_conns,
+                   int a_repeat_count, TimeNs a_repeat_interval);
+
+    void schedule() override final;
```
PaulRalnikov, line 15:
```suggestion
    void schedule() final;
```

source/scenario/action/send_data_action.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+private:
+    TimeNs m_when;
+    SizeByte m_size;
+    std::vector<std::weak_ptr<IConnection>> m_conns;
+    int m_repeat_count;
```
PaulRalnikov, line 21:
```suggestion
    std::size_t m_repeat_count;
```

source/simulator.hpp
```diff
@@ -49,6 +51,7 @@ class Simulator {
     std::unordered_set<std::shared_ptr<ISwitch>> m_switches;
     std::unordered_set<std::shared_ptr<IConnection>> m_connections;
     std::unordered_set<std::shared_ptr<ILink>> m_links;
+    std::shared_ptr<Scenario> m_scenario;
```
PaulRalnikov, line 54: I think we do net need any pointer here because `Scenario` is a concrete object, not abstract
```suggestion
    Scenario m_scenario;
```

PaulRalnikov (changes requested):

PR #467, opened 2025-10-13
BUG: Fixing a bug with multiple conns

Az3git:


scripts/generators/topology/fractal/fractal.py
```diff
@@ -3,6 +3,14 @@
 from common import *
 from generators.topology.common import *
 
+def _receiver_names(cfg: dict) -> list[str]:
+    n = int(cfg.get("receivers", 1))
+    if n < 1:
+        raise ValueError("Config 'receivers' must be >= 1")
+    if n == 1:
+        return ["receiver"]
+    return [f"receiver-{i}" for i in range(1, n + 1)]
```
PaulRalnikov: Lets make code more clear
  
    
      
        Suggested change
      
    
    
      
          
            
            def _receiver_names(cfg: dict) -> list[str]:
          
          
            
                n = int(cfg.get("receivers", 1))
          
          
            
                if n < 1:
          
          
            
                    raise ValueError("Config 'receivers' must be >= 1")
          
          
            
                if n == 1:
          
          
            
                    return ["receiver"]
          
          
            
                return [f"receiver-{i}" for i in range(1, n + 1)]
          
          
            
            def generate_receiver_names(receivers_count : int) -> list[str]:
          
          
            
                if receivers_count == 1:
          
          
            
                    return ["receiver"]
          
          
            
                return [f"receiver-{i}" for i in range(1, receivers_count + 1)]

scripts/generators/topology/fractal/default_config.yml
```diff
@@ -10,4 +10,5 @@ packet-spraying:
   type: ecmp
 
 depth: 3 # count of layers
-num_switches_per_device: 2 # count of childs of each device
\ No newline at end of file
+num_switches_per_device: 2 # count of childs of each device
+receivers: 1 # count of receivers, must be >= 1
```
PaulRalnikov: Suggestion:
```suggestion
receivers_count: 1 # count of receivers, must be >= 1
```

source/connection/flow/tcp/tcp_flow.cpp
```diff
@@ -295,24 +301,30 @@ std::string TcpFlow::to_string() const {
...
 
-TimeNs TcpFlow::get_max_timeout() const {
-    std::optional<TimeNs> mean = m_rtt_statistics.get_mean();
-    if (!mean.has_value()) {
-        return TimeNs(std::numeric_limits<double>::max());
+// Before the first ACK: exponential growth by timeout
+void TcpFlow::update_rto_on_timeout() {
+    if (!m_rto_steady) {
+        TimeNs doubled_rto = m_current_rto * 2;
+        m_current_rto = (doubled_rto > m_max_rto) ? m_max_rto : doubled_rto;
```
PaulRalnikov: Might be simplified
```suggestion
        m_current_rto = std::min(m_current_rto * 2, m_max_rto);
```

PaulRalnikov (changes requested):

PaulRalnikov (approved):
LGTM!

PR #469, opened 2025-10-15
REFACTORING: summary refactoring

Az3git:


source/connection/flow/i_flow.hpp
```diff
@@ -19,6 +19,8 @@ class IFlow : public Identifiable {
     virtual SizeByte get_sending_quota() const = 0;
     virtual std::optional<TimeNs> get_last_rtt() const = 0;
     virtual SizeByte get_delivered_data_size() const = 0;
+    virtual SizeByte get_sent_data_size() const = 0;
+    virtual uint32_t retransmit_count() const = 0;
```
PaulRalnikov: I think total size of retransmitted data is more representative
```suggestion
    virtual SizeByte retransmit_count() const = 0;
```

source/utils/summary.cpp
```diff
@@ -4,19 +4,43 @@
...
+            const SizeByte delivered = flow->get_delivered_data_size();
+            real_data_delivery += delivered;
+
+            const SizeByte sent = flow->get_sent_data_size();
+            if (sent < delivered) {
+                throw std::runtime_error(fmt::format(
+                    "For flow {} of connection {} sent {} bytes but delivered "
+                    "{} bytes",
+                    flow->get_id(), conn_id, sent.value(), delivered.value()));
+            }
```
PaulRalnikov: I think its not good idea to throw error here: you can construct summary, its correct table, but its content is unexpected. In my opinion its better to create `std::optional<ErrorMessage> Summary::check() const` (where `ErrorMessage` is the alias for `std::string`) and call it after constructing Summary
```suggestion
```

source/utils/summary.cpp
```diff
@@ -4,19 +4,43 @@
...
+
+            m_values[conn_id][flow->get_id()] =
+                FlowSummary{sent,       delivered, sending_rate,
+                            throughput, fct,       retransmit_count};
+        }
+        if (expt_data_delivery > real_data_delivery) {
+            throw std::runtime_error(fmt::format(
+                "For connection {} expected delivery {} but real is {}",
+                conn_id, expt_data_delivery.value(),
+                real_data_delivery.value()));
```
PaulRalnikov: Same

source/utils/summary.cpp
```diff
@@ -1,39 +1,84 @@
...
+                << ", " << fs.throughput.value() << ", " << fs.fct.value()
+                << "\n";
         }
     }
     out.close();
 }
 
+void Summary::check() const {
+    if (!m_errors.empty()) {
+        throw std::runtime_error(fmt::format("{}", fmt::join(m_errors, "\n")));
```
PaulRalnikov:
```suggestion
        throw std::runtime_error(fmt::format("Summary errors: {}", fmt::join(m_errors, "\n")));
```

test/simulator/start_test.cpp
```diff
@@ -12,7 +12,11 @@ namespace test {
 
 namespace {
 
-std::shared_ptr<sim::TcpFlow> add_connection_with_single_flow(
+std::shared_ptr<sim::TcpFlow> add_connection_with_single_flow( 
+
+
+
+    
     sim::Simulator& sim, const Id& conn_id,
```
PaulRalnikov: Rollback it
```suggestion
std::shared_ptr<sim::TcpFlow> add_connection_with_single_flow( 
    sim::Simulator& sim, const Id& conn_id,
```

PR #473, opened 2025-10-23
FEAT: path hash in packet

### LINKED ISSUE #474: FEAT: path hash in packet
Needs to add new field to packet that represents hash of route packet passed

PaulRalnikov:
closes #474


test/switch/test_switch.cpp
```diff
@@ -164,4 +165,55 @@ TEST_F(TestSwitch, test_one_sender) { test_senders(1); }
 
 TEST_F(TestSwitch, test_multiple_senders) { test_senders(5); }
```
Az3git: Add a small comment about what is happening in this test - it may be forgotten in the future

PR #482, opened 2025-11-14
FEAT: adds conn and mplb mocks

Az3git:


source/connection/flow/tcp/i_tcp_cc_factory.hpp
```diff
@@ -7,4 +9,4 @@ class ITcpCCFactory {
     virtual std::unique_ptr<ITcpCC> create_cc() = 0;
 };
 
-}  // namespace sim
\ No newline at end of file
+}  // namespace sims
```
PaulRalnikov: Why not Sims 4?
```suggestion
}  // namespace sim
```

PR #492, opened 2026-01-05
Implementation universal presets

### LINKED ISSUE #484: Implement universal presets
Currently topology parser supports presets only for links. Needs to add support for all types of objects. Supposed plan:
1. Change API & realization of objects parsers: currently most of them takes one `ConfigNode` (maybe with some additional info) that they should parse into concrete object. In new implementation parsers should take two nodes: main node and preset node (last one should be used if some field is missing in main node). Also it might be useful to wrap preset node with `std::optional`
2. Replace old way of link presets with new one 
3. Add preset for flows in simulation configs

S1mpotyaga:
close #484

### DIFF
--- a/source/parser/config_reader/config_node_with_preset.cpp
+++ b/source/parser/config_reader/config_node_with_preset.cpp
@@ -8,9 +8,9 @@ ConfigNodeWithPreset::ConfigNodeWithPreset(ConfigNode a_node, std::optional<Conf
 ConfigNodeWithPresetExpected ConfigNodeWithPreset::operator[](std::string_view key) const{
     const ConfigNodeExpected child_node = m_node[key];
     if (!child_node.has_value()){
-    // m_node hasn't key. Tries to find key in m_preset
+        // m_node hasn't key. Tries to find key in m_preset
         if (!m_preset.has_value()){
-        // m_preset doesn't exist. Tries to find preset for this node
+            // m_preset doesn't exist. Tries to find preset for this node
             if (!m_presets_node.has_value()){
                 // m_presets_node doesn't exist. Returns message about it
                 std::stringstream ss;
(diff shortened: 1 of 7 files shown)


source/parser/config_reader/config_node_with_preset.cpp
```diff
@@ -0,0 +1,52 @@
...
+                ss << "Error: preset named" << preset_name << "of node" << m_node << '\n';
+                ss << "can not be found in presets node:" << presets_node << ":\n";
+                return std::unexpected(ss.str());
+            }
+            m_preset.emplace(preset_node.value());
+        }
+        ConfigNodeExpected key_node = m_preset.value()[key];
+        if (!key_node.has_value()){
+            std::stringstream ss;
+            ss << "Key error: preset " << m_preset.value().get_name().value() << " ";
```
Az3git:
```suggestion
            ss << "Key error: preset '" << m_preset.value().get_name().value() << "' ";
```

source/parser/config_reader/config_node_with_preset.cpp
```diff
@@ -0,0 +1,52 @@
...
+            }
+            ConfigNodeExpected node_preset_name = m_node["preset-name"];
+            if (!node_preset_name.has_value()){
+                std::stringstream ss;
+                ss << "Key error: node " << m_node << ":\n";
+                ss << "does not have key `preset-name`";
+                return std::unexpected(ss.str());
+            }
+            ConfigNode presets_node = m_presets_node.value();
+            std::string preset_name = node_preset_name.value().as_or_throw<std::string>();
```
Az3git: Maybe better to have as_expected<std::string>() or as<std::string>() return expected, and wrap the error in unexpected

source/parser/config_reader/config_node_with_preset.hpp
```diff
@@ -0,0 +1,45 @@
...
+    template <typename T>
+    [[nodiscard]] utils::StrExpected<T> as() const noexcept {
+        try {
+            return m_node.as<T>();
+        } catch (const std::exception& e) {
+            return std::unexpected(e.what());
+        } catch (...) {
+            return std::unexpected(
+                "Undefined exception while calling `YAML::Node::as`");
+        }
```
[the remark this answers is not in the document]
PaulRalnikov: Yes, you are right

PR #504, opened 2026-02-08
method to_string for all units

### LINKED ISSUE #501: FEAT: `to_string` method for all units
Needs to add `to_string` method to all units (`Time`, `Speed`, `Size`). This method should return rounded (if needed) value with corresponding suffix (`ns`, `ms`, `Gbps` etc) **that also used in parsing** of this unit. 

Probably its good time to add this suffixes to unit bases. For example change `Bit` like from:

```
struct Bit {
    static constexpr uint64_t to_bit_multiplier = 1;
};
````

to:

```
struct Bit {
    static constexpr uint64_t to_bit_multiplier = 1;
    statinc constexpr std::string_view suffix = "b";
};
```

And change corresponding concept from:
```
// Concept for all types that might be the base for Size
template <typename T>
concept IsSizeBase = requires {
    { T::to_bit_multiplier } -> std::convertible_to<uint64_t>;
};
```
to:
```
// Concept for all types that might be the base for Size
template <typename T>
concept IsSizeBase = requires {
    { T::to_bit_multiplier } -> std::convertible_to<uint64_t>;
    { T::suffix } -> std::convertible_to<std::string_view>;
};
```

### PR DESCRIPTION
**Pull request by S1mpotyaga**
close #501

### CODE REVIEW

File: source/units/size.hpp [hunk 1]
```diff
@@ -7,40 +8,49 @@
 
 struct Bit {
     static constexpr uint64_t to_bit_multiplier = 1;
+    static constexpr std::string_view suffix = "b";
```
PaulRalnikov: Use this suffixes in units parsing. See `sourse/parser/parse_utils.cpp`

source/units/speed_fmt.hpp
```diff
@@ -0,0 +1,56 @@
...
+template<IsSizeBase TSizeBase, IsTimeBase TTimeBase>
+class fmt::formatter<Speed<TSizeBase, TTimeBase>>{
+    
+    // Parses precision which insides in {}. Precision must be specified as follows: {:3}.
+    constexpr auto parse(fmt::format_parse_context& context){
+        auto it = context.begin();
+        auto end_ptr = contex.end();
+        int rounding = 0;
+        bool is_meet_colon = false; 
+        short is_meet_braces = 0; // the first bit indicates that '{' met. the second bit indicates that '}' met.
```
PaulRalnikov: Please, use more simple construction like `bool is_meet_left_brace`, `bool is_meet_right_brace`. There is no point to complicate code like that

source/parser/parse_utils.hpp
```diff
@@ -8,7 +8,7 @@
 
 #include "logger/logger.hpp"
 #include "parser/config_reader/config_node.hpp"
-#include "types.hpp"
+#include "../types.hpp"
```
PaulRalnikov: No need; see cmake file for more info
```suggestion
#include "types.hpp"
```

source/units/common_formatter.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <spdlog/fmt/bundled/format.h>
+
+template<typename Char>
+struct CommonFormatter: fmt::formatter<double, Char>{
```
PaulRalnikov: 1) I think this class might be simple done more common: instead of using concrete base formatter `fmt::formatter<double, Char>` abstract by formatting type
2) Use speaking name like `DefaltFormatter`
```suggestion
template<typename T, typename Char>
struct CommonFormatter: fmt::formatter<T, Char>{
```

and please, check that `fmt::formatter` is implemented for type `T` using concepts. `fmt` already containts concept you need -- `fmt::formattable<T, Char>`, so all you need is just to use it

source/units/size_fmt.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <spdlog/fmt/bundled/format.h>
+#include <cstdlib>
+
+#include "size.hpp"
+#include "common_formatter.hpp"
+
+template <IsSizeBase TSizeBase, typename Char>
+struct fmt::formatter<Size<TSizeBase>, Char> : CommonFormatter<Char> {
```
PaulRalnikov: Here you use double formatter as base for `CommonFormatter` that is strange - `Size<TSizeBase>` store value as `uint64_t`

source/units/size_fmt.hpp
```diff
@@ -0,0 +1,28 @@
...
+            out = fmt::formatter<double, Char>::format(t.value(), ctx);
+        }
+        for (auto c : TSizeBase::suffix) {
+            *out++ = static_cast<Char>(c);
+        }
+        return out;
+    }
+
+private:
+    static constexpr const char* default_precision_format_str = "{}";
```
PaulRalnikov: Use `std::string_view` instead of `const char*`
```suggestion
    static constexpr std::string_view default_precision_format_str = "{}";
```

source/units/speed.hpp
```diff
@@ -71,6 +72,16 @@ class Speed {
         return !(*this == speed);
     }
 
+    // Outputs the Speed with the specified precision and with suffix.
+    // For example: 3.35 Mb/ms
+    std::string constexpr to_string(int precision = 2) const{
+        std::stringstream ss;
+        ss << std::fixed << std::setprecision(precision) << value() << TSizeBase::suffix << '/' << TTimeBase::suffix;
```
PaulRalnikov: Use separator instead of hardcoded `'/'`
```suggestion
        ss << std::fixed << std::setprecision(precision) << value() << TSizeBase::suffix << m_separator << TTimeBase::suffix;
```

source/units/speed_fmt.hpp
```diff
@@ -0,0 +1,32 @@
...
+        if (this->use_default_precision) {
+            out = fmt::format_to(out, this->default_precision_format_str, t.value());
+        } else {
+            out = fmt::formatter<double, Char>::format(t.value(), ctx);
+        }
+        for (auto c : TSizeBase::suffix) {
+            *out++ = static_cast<Char>(c);
+        }
+        auto sep = Speed<TSizeBase, TTimeBase>::m_separator;
+        out = std::copy(sep.begin(), sep.end(), out);
```
PaulRalnikov: Looks strange - in this line you use `std::copy`, but in next one use `for`. Choose only one
```suggestion
        for (auto c : Speed<TSizeBase, TTimeBase>::m_separator) {
            *out++ = static_cast<Char>(c);
        }
```

source/units/time_fmt.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include "time.hpp"
+#include "common_formatter.hpp"
+
+template <IsTimeBase TTimeBase, typename Char>
+struct fmt::formatter<Time<TTimeBase>, Char> : CommonFormatter<Char> {
+
+    template <typename FormatContext>
+    auto format(const Time<TTimeBase>& t, FormatContext& ctx) const {
+        auto out = ctx.out();
+        if (this->use_default_precision) {
```
PaulRalnikov: Same about `this->`
```suggestion
        if (use_default_precision) {
```

test/simulator/format/size_format_test.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "units/size_fmt.hpp"
+
+class SizeFormatTest: public testing::Test{
+public:
+    void TearDown() override {};
+    void SetUp() override {};
+};
+
+TEST_F(SizeFormatTest, DefaultValue) {
```
PaulRalnikov: In case of empty class you may not create it explicitly, just use `TEST` macros instead of `TEST_F`. In such case googletest create empty class automaticly
```suggestion

TEST(SizeFormatTest, DefaultValue) {
```

source/parser/parse_utils.cpp
```diff
@@ -45,10 +45,14 @@ utils::StrExpected<SpeedGbps> parse_speed(const std::string &throughput) {
...
+    std::string_view separator_gbit_second = Speed<GBit, Second>::m_separator;
+    std::string check_suffix_mbps = std::string(MBit::suffix) + std::string(separator_mbit_second) + std::string(Second::suffix);
+    std::string check_suffix_gbps = std::string(GBit::suffix) + std::string(separator_gbit_second) + std::string(Second::suffix);
+    if (unit == check_suffix_mbps) {
         return SpeedGbps(Speed<MBit, Second>(value));
     }
-    if (unit == "Gbps") {
+    if (unit == check_suffix_gbps) {
         return SpeedGbps(value);
     }
```
PaulRalnikov: Looks strange
1) Two equal fields with different names
2) It might be simplier to create static `static constexpr std::string get_suffix()` method for `Speed` and use it here

```suggestion
    if (unit == Speed<Mbit, Second>::suffix()) {
return SpeedGbps(Speed<MBit, Second>(value));
}
    if (unit == Speed<Gbit, Second>::suffix()) {
return SpeedGbps(value);
}
```

